PR #51, opened 2012-09-24
Exported resources

hunner:
This pull request allows rspec-puppet module spec tests to be written for classes which contain the exported resource collector `<<| |>>` by defining `let(:exported_resources)` arrays of resource hashes. It builds off of @rodjek's previous work, with some modifications to remove monkey-patching and work around rspec bugs.

This currently implements exporting resources for importing. It does not have the ability to check that resources were exported by the catalogue compile because the temporary database is wiped after each test case (but rspec-puppet caches the catalogue instead of compiling for each test case).

There is currently no way to inject virtual resources into a catalogue compile, so this functionality has not been carried on from the original commit by @rodjek.

This change is backwards compatible. Adds 26 spec tests and all tests pass.


cyberious:
@hunner @rodjek anybody willing to pick this back up, is there an interest in being able to mock this up for anybody else?

PR #176, opened 2014-02-21
Fixes bug when spec/spec.ops missing

petems:
Previous code would cause the issue:
````
rake aborted!
No such file or directory - spec/spec.opts
```

With this change, if the file doesn't exist, the opts will be empty, and the rake will run as expected.


DavidS:
Please merge this.

PR #177, opened 2014-02-21
Adds Gemspec to Gemfile

petems:
Before this, if you wanted to check local changes to gem versions and the like using bundler, it would complain it could not find the gem in the bundle environment.


DavidS:
This looks like an easy win.

PR #186, opened 2014-03-26
Fix for issue 185.

krancour:
This will recompile the catalog when a different :hiera_config value is detected (much the same as is done when different facter facts are used).  When this condition is detected, Puppet's internally cached instance of Hiera is also cleared so that the catalog can be recompiled against the differently configured Hiera.

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -13,7 +13,9 @@ module RSpec::Puppet
       code = [import_str, pre_cond, test_manifest(type)].join("\n")
       node_name = nodename(type)
 
-      catalogue = build_catalog(node_name, facts_hash(node_name), code)
+      hiera_config_value = self.respond_to?(:hiera_config) ? hiera_config : nil
+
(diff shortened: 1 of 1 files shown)


lib/rspec-puppet/support.rb
```diff
@@ -13,7 +13,9 @@ def load_catalogue(type)
       code = [import_str, pre_cond, test_manifest(type)].join("\n")
       node_name = nodename(type)
 
-      catalogue = build_catalog(node_name, facts_hash(node_name), code)
+      hiera_config_value = self.respond_to?(:hiera_config) ? hiera_config : nil
+
+      catalogue = build_catalog(node_name, facts_hash(node_name), hiera_config_value, code)
```
DavidS: The trick seems to be that adding the configuration as additional argument will change the cache key used in build_catalog(). Without this, the cached catalog is returned unchanged.

DavidS:
I could reproduce the problem local, setting :hiera_config seems to leak across specs. I can also attest to the fact that the fix works for me with puppet 3.8.1 and puppet 4.

Moving the second chunk to `before_each_test()` works, but still requires the first chunk to get the local caching right.

PR #208, opened 2014-06-25
Set Puppet[:libdir] correctly if Puppet[:modulepath] contains multiple paths

christianberg:
When the `Puppet[:modulepath]` contains several paths (separated by `:`), the `Puppet[:libdir]` is currently not set at all, causing e.g. custom functions not to be loaded.

With this patch, the `lib` subdirectories of modules from all `modulepath` entries are added to the `libdir`.

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -133,7 +133,9 @@ module RSpec::Puppet
         end
       end
 
-      Puppet[:libdir] = Dir["#{Puppet[:modulepath]}/*/lib"].entries.join(File::PATH_SEPARATOR)
+      Puppet[:libdir] = Puppet[:modulepath].split(File::PATH_SEPARATOR).map { |d|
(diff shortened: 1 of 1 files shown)


DavidS:
Puppet only evaluates the first components `libdir`, therefore this will not work.

joshcooper:
I've submitted PR #237 to address this issue.

Prior to Puppet 4, the `libdir` could sometimes be set to an array of directories, and sometimes only to a single directory. The behavior was changed in 4 so that `libdir` must now always be a single directory, see https://tickets.puppetlabs.com/browse/PUP-3336.

Also note that adding paths to `libdir` only modifies the load path that the autoloader uses to resolve custom types, providers, functions, etc, but does not work when that code `require`s helper/utility code also in your module. This is why we need to add each module's `lib` dir to `$LOAD_PATH` instead.

DavidS:
PR #237 works for me.

PR #209, opened 2014-06-26
Allow tests to have strict variables

hunner:
Similar to how the future parser ability works, this will allow catalogs
to be compiled with strict variable parsing turned on.


DavidS:
This is superseded by #212

PR #212, opened 2014-07-09
Add support for puppet3.6 config options trusted_node_data, ordering, st...

jantman:
...ringify_facts

This effectively follows up on https://github.com/rodjek/rspec-puppet/pull/183 and the not-yet-merged https://github.com/rodjek/rspec-puppet/pull/209 by adding support for the rest of the parser-altering settings added in 3.6, specifically the three listed as "recommended and safe" (http://docs.puppetlabs.com/puppet/3.6/reference/config_important_settings.html#getting-new-features-early). ``ordering`` could prove quite useful, but the important part here is ``trusted_node_data`` which actually introduces two reserved variables, ``$facts`` and ``$trusted`` (see http://docs.puppetlabs.com/puppet/latest/reference/lang_facts_and_builtin_vars.html). This will prove valuable for modules (perhaps it should even be set to true by default?), since as it stands now, puppet4 will introduce these variables and likely break any modules that happen to use them.

### DIFF
--- a/lib/rspec-puppet.rb
+++ b/lib/rspec-puppet.rb
@@ -23,6 +23,9 @@ RSpec.configure do |c|
   c.add_setting :default_facts, :default => {}
   c.add_setting :hiera_config, :default => '/dev/null'
   c.add_setting :parser, :default => 'current'
+  c.add_setting :trusted_node_data, :default => false
+  c.add_setting :ordering, :default => 'title-hash'
+  c.add_setting :stringify_facts, :default => true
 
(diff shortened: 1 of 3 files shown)


DavidS:
Hi, would you mind rebasing this to current master and re-push to trigger another travis-ci run?

DavidS:
I've rebased this now, combined it with #209, and fixed the tests to create the new PR #293

DavidS:
This was merged in https://github.com/rodjek/rspec-puppet/pull/293 ; please close this.

PR #214, opened 2014-07-21
Add basic support for Coveralls reports

raphink:
Note:
- Requires the 'coveralls' gem
- Reports lines that have created a resource that has been asserted

This might look a bit weird in coveralls, because this is not what is usually reported by coverage, but I'm just building on top of what coverage support in rspec-puppet currently does, which is reporting resources that were declared but not tested (vs resources that were never declared when parsing the code, which is what is traditionally done).

An example of coveralls output can be found at https://coveralls.io/r/raphink/puppet-lvm

### DIFF
--- a/lib/rspec-puppet/coverage.rb
+++ b/lib/rspec-puppet/coverage.rb
@@ -5,7 +5,7 @@ module RSpec::Puppet
 
     class << self
       extend Forwardable
-      def_delegators :instance, :add, :cover!, :report!, :filters
+      def_delegators :instance, :add, :cover!, :report!, :filters, :coveralls!
     end
 
     include Singleton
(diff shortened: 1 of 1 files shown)


DavidS:
@raphink can you please review this, #219 and https://github.com/rodjek/rspec-puppet/blob/master/lib/rspec-puppet/coverage.rb ?

raphink:
@davids I'm a bit frustrated with all coverage PRs (including mine) bth. They are not *really* coverage tests. They test if there has been an assertion on each resource in the catalog, which is a bit overkill, and not really the point of coverage in the first place.

I'm afraid though that the current state of the Puppet language will not let us plug in and retrieve real coverage info (as in, which percentage of code has actually been traversed/branched in).

kylog:
@raphink this is no immediate help but just as an FYI: we started thinking through how we would add coverage to the Puppet code. There's an initial epic here: https://tickets.puppetlabs.com/browse/PUP-4764.

raphink:
Very cool @kylog, looking forward to this. In the meanwhile, I'm wondering if the current efforts here are really worth it.

PR #220, opened 2014-08-10
display_name is deprecated in RSpec

garethr:
Deprecated in 2.99 and removed in 3.0 according to https://github.com/rspec/rspec-core/blob/master/Changelog.md


DavidS:
This change was already applied with a different PR. spec-puppet works with spec 3

PR #222, opened 2014-09-05
Skip 'import' statements when parser = future

### LINKED ISSUE #196: 3.6.0 future parser removes import from language
With 3.6.0 out the door the future parser has removed support for the `import` statement. This means that if you run with the future parser on, `c.parser='future'`, all your tests turn into a sea of bloody red.

Now I'm guessing the issue is stemming from https://github.com/rodjek/rspec-puppet/blob/master/lib/rspec-puppet/support.rb#L37-L45 and the way it creates that catalog, unfortunately I have no clue how to fix this.

@jeffmccune @rodjek: Could you point me in the right direction?

3flex:
As mentioned in https://github.com/rodjek/rspec-puppet/issues/196#issuecomment-54631457, this looks to fix rspec-puppet when the future parser is used. I tested a module using Puppet 3.2, 3.5 and 3.7, and since the default is to leave behavior default when `parser != future` this shouldn't affect any existing tests when using the default parser.

If spec tests are required I'll need some help; I'm not sure how best to test the `load_catalogue` method that the code has updated.

Fixes #196

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -10,7 +10,12 @@ module RSpec::Puppet
     def load_catalogue(type)
       vardir = setup_puppet
 
-      code = [import_str, pre_cond, test_manifest(type)].join("\n")
+      if Puppet[:parser] == 'future'
+        code = [pre_cond, test_manifest(type)].join("\n")
+      else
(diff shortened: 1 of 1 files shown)


bcarpio:
@jeffmccune I didn't see an issue in the backlog for this breaking host specs. Possibly I was searching for the wrong terms.

PR #228, opened 2014-10-20
Reduce file content on summary

javiplx:
When the file content fails the check, the whole content of both files is shown in the summary line. This PR makes that only their leading lines get shown, to make the summary readable.

### DIFF
--- a/lib/rspec-puppet/errors.rb
+++ b/lib/rspec-puppet/errors.rb
@@ -11,10 +11,18 @@ module RSpec::Puppet
       end
 
       def message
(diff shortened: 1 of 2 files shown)


DavidS:
Hi Javier,

I'd like to merge this, but the five commits for a simple fix would clutter the git history. Would you mind squashing all those commits into a single one, like described here: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html ?

Regards

DavidS:
Merged that in #305.

PR #230, opened 2014-10-30
travis: Trim down the matrix, simplify versions.

daenney:
* Throw away anything relating to Puppet 2.6
* Build on 2.7 but not on too modern Ruby versions
* Build on 3.2, 3.4 and 3.6 because of PE
  * PE 3.0 -> OSS 3.2
  * PE 3.1 -> OSS 3.4
  * PE 3.3 -> OSS 3.6
* Build on latest OSS, 3.7.x
  * PE 3.7 -> OSS 3.7

By using the pessimistic version operator, `~>` we can simply specify `2.7.0` and be sure that we'll get the latest version in the 2.7.X series.

Ruby 1.9.2 was never supported by Puppet Labs, no need to test for it and Puppet 2.7 was never supported on Ruby 1.9.3+, as per: https://docs.puppetlabs.com/guides/platforms.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: ruby
 
 rvm:
   - 1.8.7
-  - 1.9.2
   - 1.9.3
   - 2.0.0
   - 2.1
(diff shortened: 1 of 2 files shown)


logicminds:
I still need 3.2.x supported.  Can we add 3.2.x back into the matrix.

.travis.yml
```diff
@@ -2,55 +2,31 @@ language: ruby
 
 rvm:
   - 1.8.7
-  - 1.9.2
   - 1.9.3
   - 2.0.0
-  - 2.1
+  - 2.1.3
```
elyscape: Why 2.1.3?  Puppet 4 ships with 2.1.5, doesn't it?

daenney: That should be 2.1 plain I think as that will always pick the latest 2.1 which is what we want.

daenney:
This is pending #273.

PR #237, opened 2014-12-03
Set $LOAD_PATH instead of puppet's libdir

joshcooper:
Commit 7f7e74f02c set puppet's `libdir` to include the lib directory for
each module in the `modulepath`. Doing so enabled puppet's autoloader to
load modules that hadn't been pluginsynced. However, this only worked for
code that the autoloader directly loaded. It didn't work when the
autoloaded type, provider, etc required helper code in the same module,
or different module, because the module's lib directory was not in ruby's
$LOAD_PATH.

The commit also relied on inconsistent behavior in puppet whereby sometimes
puppet assumed `libdir` was a (semi)colon separated list of directories, and
sometimes it assumed a single directory. In Puppet 4.0, puppet will
only interpret the `libdir` setting as a single directory[1].

As a result, this commit stops setting `libdir` and instead updates the
ruby $LOAD_PATH. In doing so, rspec-puppet is the "application" the uses
puppet as a "library". In that pattern, rspec-puppet is the thing that
sets the ruby $LOAD_PATH, and puppet just uses the specified $LOAD_PATH.
This is similar to how we fixed loading faces from modules[2] in commit[3].

[1] https://tickets.puppetlabs.com/browse/PUP-3336
[2] https://projects.puppetlabs.com/issues/7316
[3] https://github.com/puppetlabs/puppet/commit/4f99f25cd

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -127,6 +127,7 @@ module RSpec::Puppet
 
       if Puppet.version.to_f >= 4.0
         settings = [
+          [:modulepath, :module_path],
           [:environmentpath, :environmentpath],
           [:config, :config],
           [:confdir, :confdir],
@@ -153,8 +154,10 @@ module RSpec::Puppet
         end
       end
 
-      # This line is wrong. libdir should never be more than a single path
-      Puppet[:libdir] = Dir["#{Puppet[:modulepath]}/*/lib"].entries.join(File::PATH_SEPARATOR)
+      Dir["#{Puppet[:modulepath]}/*/lib"].entries.each do |lib|
+        $LOAD_PATH << lib
+      end
+
       vardir
     end
 
--- a/spec/support_spec.rb
+++ b/spec/support_spec.rb
@@ -18,5 +18,18 @@ describe RSpec::Puppet::Support do
       subject.setup_puppet
       expect(Puppet[:parser]).to eq("future")
     end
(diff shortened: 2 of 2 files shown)


fatmcgav:
@joshcooper Just came back to this one... 

Any more thoughts?

joshcooper:
@fatmcgav thanks for the ping. I think we need to figure out why `RSpec::Puppet::Support#setup_puppet` is never getting called. I would expect `rspec-puppet` to call it automatically as part of test framework setup so that modules don't need to, but I'm not sure about that.

Since `setup_puppet` is never called, many things are not configured correctly, e.g. `vardir` is set to `/dev/null`, in addition to the specific issue of `$LOAD_PATH` not being updated. Once we call `setup_puppet` things should just work.

DavidS:
Yeah, I've rebased and tested this locally and it works fine for me.

joshcooper:
@hunner @DavidS I've updated the PR to resolve the conflict. I updated the `setup_puppet` method so that it allows `modulepath` to be specified, which is probably not the right thing for Puppet 4. Alternatively, I could something like the following, though I'm not sure what `basedir` should be  in rspec-puppet?

```
basedir = Puppet.version.to_f >= 4.0 ? Puppet[:environmentpath] : Puppet[:modulepath]
Dir["#{basedir}/*/lib"].entries.each do |lib|
  $LOAD_PATH << lib
end
```

DavidS:
Most important for rspec-puppet, as I'm currently using it, is to have the containing module's `lib/` in the load path.

Issue #245, opened 2014-12-13
Rspec.deprecate()'s API has changed.


voidus:
The commit https://github.com/rspec/rspec-core/commit/c07c4cf9b47b9f58b222b77ca13daea11b3d7585 changed the API of RSpec.deprecate().

When I try to run `rake rspec` in https://github.com/wyrie/puppet-backuppc/commit/acf4baafd45e8b554bdda29d9481d2e0898b3713, I get the following error:
```
Failures:

  1) backuppc::client On Ubuntu should include Class[backuppc::params]
     Failure/Error: it { should include_class("backuppc::params") }
     TypeError:
       no implicit conversion of Symbol into Hash
     # /home/voyd/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/rspec-core-2.99.2/lib/rspec/core/deprecation.rb:12:in `merge'
     # /home/voyd/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/rspec-core-2.99.2/lib/rspec/core/deprecation.rb:12:in `deprecate'
     # /home/voyd/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/rspec-puppet-1.0.1/lib/rspec-puppet/matchers/include_class.rb:7:in `block (2 levels) in <module:ManifestMatchers>'
     # /home/voyd/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/rspec-expectations-2.99.2/lib/rspec/matchers/extensions/instance_eval_with_args.rb:11:in `instance_exec'
     # /home/voyd/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/rspec-expectations-2.99.2/lib/rspec/matchers/extensions/instance_eval_with_args.rb:11:in `instance_eval_with_args'
     # /home/voyd/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/rspec-expectations-2.99.2/lib/rspec/matchers/matcher.rb:99:in `matches?'
     # /home/voyd/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/rspec-expectations-2.99.2/lib/rspec/expectations/handler.rb:24:in `handle_matcher'
     # /home/voyd/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/rspec-core-2.99.2/lib/rspec/core/memoized_helpers.rb:77:in `should'
[...]

PR #246, opened 2014-12-13
Revert "Clean up include_class deprecation notice"

voidus:
If the deprecated `include_class` is used with rspec-2.99, #245 occurs.
#### Commit message of the fix
This reverts commit 8d951da86be5e61e17fec4491cb2990b8efb590b.

The RSpec-core commit c07c4cf9b47b9f58b222b77ca13daea11b3d7585
(included in 2.99.0) has changed the deprecation api:
```
    - def deprecate(deprecated, replacement_or_hash={}, ignore_version=nil)
    - data = Hash === replacement_or_hash
                        ? replacement_or_hash
                        : { :replacement => replacement_or_hash }
    + def deprecate(deprecated, data = {})
```
The commit that has been reverted changed the second parameter from
    `replacement: foo`
to
    `foo`
which is incompatible with the new form, so it has to be reverted.

Now it works with 2.X again.

Conflicts:
    lib/rspec-puppet/matchers/include_class.rb
        context conflict

The tests have also been adjusted accordingly

### DIFF
--- a/lib/rspec-puppet/matchers/include_class.rb
+++ b/lib/rspec-puppet/matchers/include_class.rb
@@ -4,7 +4,7 @@ module RSpec::Puppet
 
     matcher :include_class do |expected_class|
       match do |catalogue|
-        RSpec.deprecate(:include_class, :contain_class)
+        RSpec.deprecate(:include_class, :replacement => :contain_class)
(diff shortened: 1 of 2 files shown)


DavidS:
Merged that in #305.

PR #261, opened 2015-03-11
Improve dependency detection

raphink:
When refactoring (and potentially in other cases), it is useful to express resource dependencies that are not direct.

This PR adds two levels of recursion to dependency detection:

* test dependencies on all levels, which checks for dependencies with parents of the resource (i.e. the resource is contained, directly or not, in the class involved in the relationship). You can think of this as a vertical recursion. E.g. where Package['foo'] comes before File['/foo']:

```puppet
class { 'foo::install': } ->
class { 'foo::config': }

class foo::install {
  package { 'foo': }
}

class foo::config {
  file { '/foo': }
}
```

* test dependencies recursively by following indirect dependencies (dependencies of dependencies). This is a horizontal recursion. E.g. where Yumrepo['foo'] comes before File['/foo']:

```puppet
class { 'foo::repo': } ->
class { 'foo::install': } ->
class { 'foo::config': }

class foo::repo {
  yumrepo { 'foo': }
}

class foo::install {
  package { 'foo': }
}

class foo::config {
  file { '/foo': }
}
```

In addition to this, this series also adds support for autorequires detection. I put this there because it goes tightly with the recursive code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,49 @@ When testing custom types, the `be_valid_type` matcher provides a range of expec
 * `with_features(<feature_list>)`: check that the specified features are available
 * `with_set_attributes(<param_value_hash>)`: check that the specified attributes are set 
 
+#### Recursive dependencies
+
+The relationship matchers are recursive in two directions:
+
+* vertical recursion, which checks for dependencies with parents of the resource
+ (i.e. the resource is contained, directly or not, in the class involved in the relationship).
+ E.g. where `Package['foo']` comes before `File['/foo']`:
+
+```puppet
+class { 'foo::install': } ->
+class { 'foo::config': }
+
+class foo::install {
+  package { 'foo': }
+}
+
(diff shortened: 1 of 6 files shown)


mcanevet:
+1

mcanevet:
@jeffmccune do you have any comment on this?
We'd really like to make this work.
(related post in puppet-dev : https://groups.google.com/forum/#!topic/puppet-dev/76nGtIKNkI8)

raphink:
@zaphod42: I'm guessing you're probably the most knowledgeable person regarding this kind of code. @peiriannydd also maybe?

strider:
:+1:

igalic:
is there any way we can get the spec tests to pass? all of them, that is…

raphink:
zrqkf6j

raphink:
@DavidS rebased, squashed, documented.

PR #265, opened 2015-04-08
Adding stuff for puppet 4

hunner:
Puppet 4 uses environmentpath instead of modulepath and manifest
settings. We can use an environment of 'fixtures' to access the
spec/fixtures/{manifests,modules} directories for compiling a catalog.

### DIFF
--- a/lib/rspec-puppet.rb
+++ b/lib/rspec-puppet.rb
@@ -14,6 +14,7 @@ rescue LoadError
 end
 
 RSpec.configure do |c|
+  c.add_setting :environmentpath, :default => '/etc/puppetlabs/code/environments'
   c.add_setting :module_path, :default => '/etc/puppet/modules'
   c.add_setting :manifest_dir, :default => nil
   c.add_setting :manifest, :default => nil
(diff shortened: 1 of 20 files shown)


hunner:
Requires puppetlabs/puppetlabs_spec_helper#100

lib/rspec-puppet.rb
```diff
@@ -14,6 +14,7 @@
 end
 
 RSpec.configure do |c|
+  c.add_setting :environmentpath, :default => '/etc/puppet/environments'
```
kylog: FYI, puppet 4 is using different file paths. See https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md.

mgagne:
Can you explain how someone can test against a environment other than "fixtures"? I'm using rspec-puppet to test against Hiera data and I want to able to test against the "production" environment. (or any other environments) BTW, why not "production" by default?

hunner:
@mgagne I'm not sure how you are setting the environment to a custom setting for an rspec-puppet run, but I switched from using a static environment named after the `node_obj`'s environment rather than a hard-coded name. Let me know if this works for you!

DavidS:
This needs another Environment.new -> Environment.create fix in lib/rspec-puppet/support.rb#build_node .

hunner:
@DavidS I actually had that locally from @electrical but forgot to push it. Pushed now.

jhoblitt:
Should this be working with puppetlabs_spec_helper 0.10.2?

PR #267, opened 2015-04-15
Decouple the message from what the subject function does to it's argumen...

DavidS:
...ts

Before that, a subject that would e.g. shift the argument array would cause the
final message to be modified. With this change, the @func_args are initialised
immediately and not accessible anymore to modificatoin from the subject.


DavidS:
I should probably do add a spec for this case. It's easy to loose

igalic:
:+1:

hunner:
:+1:

PR #271, opened 2015-04-21
Unify environment setting and allow customization

hunner:
Before this commit, the environment was test, testing, or production
depending on which kind of test it was. This commit unifies it to be a
unique string that should be useful for validating tests that require
environment to be set, and documents how to set it.


strider:
:+1:

PR #273, opened 2015-04-21
Fix coverage on puppet 4

hunner:
Coverage needs to filter out fixtures, but the method of determining
what classes are declared by fixtures relied on Puppet[:manifest] and
Puppet[:modulepath] which are not used in Puppet 4.


strider:
:+1:

hunner:
Except it's not green in travisci, but I can't get it to fail locally...

3flex:
:+1: 

The failure is because of https://github.com/rodjek/rspec-puppet/commit/d1feb3919ee0579471ddb9576a24d0a258a4345a (#267), see https://travis-ci.org/rodjek/rspec-puppet/builds/59430415

hunner:
I think I got it now...

PR #274, opened 2015-04-22
Include pops from puppet4

DavidS:


spec/functions/nasty_spec.rb
```diff
@@ -1,6 +1,10 @@
 require 'spec_helper'
 
 describe 'nasty' do
+
+  it { expect(subject).not_to eq(nil) }
+  it { expect(subject).to run.with_params('foo', 'bar').and_return('foo') }
```
hunner: Doesn't this test basically just duplicate what's down below?

PR #277, opened 2015-04-28
Don't match compile.and_raise_error on success

### LINKED ISSUE #276: compile.and_raise_error passes if no error is raised
``compile.and_raise_error(...).matches?`` returns ``true`` if no error is raised during compilation. This matcher will not detect an error if an exception is _incorrectly_ not raised.

``compile.and_raise_error`` sets ``@expected_error``, but then ``matches?`` does not check if ``@expected_error`` is set before returning ``true``.

``run.and_raise_error`` is similar, but ``run.and_raise_error(...).matches?`` does have the correct behavior of returning ``false`` if an exception is not raised.

jmoldow:
Make the behavior of `compile.and_raise_error` match that of
`run.and_raise_error` and `should raise_error`: if compilation succeeds
and does not in fact raise an error, this should be considered a unit
test failure that is reported.

It would previously succeed even if no error was raised.

Fixes #276.

### DIFF
--- a/lib/rspec-puppet/matchers/compile.rb
+++ b/lib/rspec-puppet/matchers/compile.rb
@@ -27,7 +27,7 @@ module RSpec::Puppet
           elsif @check_deps == true && missing_dependencies?
             false
           else
-            true
+            @expected_error.nil?
           end
         rescue Puppet::Error => e
           @error_msg = e.message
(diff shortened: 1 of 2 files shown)


DavidS:
Excellent catch. Thanks for great PR with tests!

PR #280, opened 2015-04-30
Allow undef parameter passing with nil

### LINKED ISSUE #279: Can't pass undef as a param value
When testing a definition or parametrized class, it isn't possible to use ``let(:params)`` to set one of the values to Puppet's ``undef`` value.

One might try doing something like ``let(:params) { { :foo => nil } }``, but this doesn't work. When constructing the declaration, the value is inspected to a string, which in this case is ``'nil'``. The resource declaration that is constructed is ``class { 'test': foo => nil }``, which is not valid Puppet and caused an error to be raised.

Doing something like ``let(:params) { { :foo => :undef } }`` is similarly invalid. And doing something like ``let(:params) { { :foo => 'undef' } }`` leads to the resource declaration ``class { 'test': foo => "undef" }``, which is valid but doesn't actually use the special undef value.

When ``nil`` is used as a param value, it should not be inspected, but should instead lead to the string ``'undef'`` being used in the generated resource declaration.

Another solution, proposed at the end of [1], would be to define a special ``Undef`` class whose ``inspect`` method returns the unquoted string ``'undef'``. Then ``let(:params) { { :foo => Undef.new } }`` leads to the desired resource declaration ``class { 'test': foo => undef }``.

I believe the first solution is better and more developer friendly, because it uses the Ruby native ``nil`` value and doesn't introduce any extra types that developers need to know about. And after all, Puppet ``undef`` is nearly equivalent to Ruby ``nil``, so it makes sense to do this translation.

[1] https://groups.google.com/forum/#!topic/puppet-users/6nL2eROH8is

jmoldow:
Now, when `nil` is used as a param value, its `inspect` method is not
called to construct the Puppet parameter value. Instead, the unquoted
string `'undef'` is used in the generated resource declaration. So now
`let(:params) { { :foo => nil } }` leads to the desired resource
declaration `class { 'test': foo => undef }`.

Puppet `undef` is nearly equivalent to Ruby `nil`, and there is no `nil`
in Puppet, so it makes sense to do this translation.

Before this change, it was impossible to pass `undef` as a parameter.
Using `nil`, `:undef`, or `'undef'` all failed for various reasons.

Credit to the discussion at [1] for identifying the issue and proposing
an alternate solution.

Fixes #279.

[1] https://groups.google.com/forum/#!topic/puppet-users/6nL2eROH8is

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,12 @@ let(:title) { 'foo' }
 let(:params) { {:ensure => 'present', ...} }
 ```
 
+##### Passing Puppet's `undef` as a parameter value
+
+```ruby
+let(:params) { {:user => nil, ...} }
+```
+
 #### Specifying the FQDN of the test node
 
 If the manifest you're testing expects to run on host with a particular name,
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -116,8 +116,10 @@ module RSpec::Puppet
 
     def param_str
       params.keys.map do |r|
-        param_val = escape_special_chars(params[r].inspect)
-        "#{r.to_s} => #{param_val}"
+        param_val = params[r]
+        is_undef = param_val.nil?
+        param_val_str = escape_special_chars(is_undef ? 'undef' : param_val.inspect)
+        "#{r.to_s} => #{param_val_str}"
       end.join(', ')
     end
 
--- a/spec/classes/undef_spec.rb
+++ b/spec/classes/undef_spec.rb
@@ -1,0 +1,18 @@
+require 'spec_helper'
+
+describe 'undef' do
+  it { should compile.with_all_deps }
+
+  shared_examples 'exec echo' do
+    it { should contain_exec('/bin/echo foo').with_user(nil) }
+  end
+
+  context 'with user => nil' do
+    let(:params) { { :user => nil } }
+    include_examples 'exec echo'
+  end
+
+  context 'with params unset' do
+    include_examples 'exec echo'
+  end
+end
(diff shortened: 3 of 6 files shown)


igalic:
interestingly, it appears that your tests are failing.

jmoldow:
Thanks @igalic . My tests are failing because I named my test class ``undef``, which is an invalid identifier name. I renamed it locally, and the tests pass. I just haven't pushed the fix yet.

Though I'm questioning whether this is the correct approach. I was looking into the code for the generic matchers, and I saw that ``nil`` has a different meaning there. When doing any of ``.with_foo``, ``.with_foo(nil)``, ``.with({:foo=>nil})``, ``.with([:foo,:bar])``, the expected value of ``:foo`` is taken to be ``nil``, but this doesn't mean ``undef``, but means any value. Similarly, when doing any of `.without_foo``, ``.without_foo(nil)``, ``.without({:foo=>nil})``, ``.without([:foo,:bar])``, the unexpected value of ``:foo`` is taken to be ``nil``, but this doesn't mean not ``undef``, but means not any value.

Since not passing a parameter is not always semantically equivalent to passing an ``undef`` value, we can't just use ``.without_foo`` to test for ``undef`` being passed as the value for ``:foo``.

So it might make sense to go with the suggestion in https://groups.google.com/forum/#!topic/puppet-users/6nL2eROH8is, and introduce a special ``Undef`` class, which can be used both with ``let(:params)`` and with ``.with`` and ``.without``.

@igalic, what do you think?

DavidS:
I like the idea with a special `Undef` class. @jmoldow are you still interested in getting this cleaned up and merged?

DavidS:
This was merged as #335 . Thanks to all involved for getting a good solution in!

domcleal:
I think this PR can be closed now.

PR #281, opened 2015-05-01
FunctionMatchers::Run: make return value optional

DavidS:
I'm not sure whether that's a good solution, but without this, functions like `validate_...` have many descriptions with a dangling 'and'. `@desc` could be initialised to `'return nothing'`by default to avoid having to specify `.and_return` in all the specs.


DavidS:
Scratch this. `@expected_result` being `nil` is used to signal to fully turn off result checking, creating a whole lot of different problems.

PR #283, opened 2015-05-04
Several bug fixes and specs for Run matcher

DavidS:
Without the changes to run.rb, the following specs would fail:

```
rspec ./spec/unit/matchers/run_spec.rb:10 # RSpec::Puppet::FunctionMatchers::Run should not match a lambda that raises an error

rspec ./spec/unit/matchers/run_spec.rb:24 # RSpec::Puppet::FunctionMatchers::Run with_params() should not match a lambda that raises an error

rspec ./spec/unit/matchers/run_spec.rb:40 # RSpec::Puppet::FunctionMatchers::Run with_params(*anything*) and_return(*anything*) should not match a lambda that raises an error

rspec ./spec/unit/matchers/run_spec.rb:36 # RSpec::Puppet::FunctionMatchers::Run with_params(*anything*) and_return(nil) should not match a lambda that does return a different value
```

### DIFF
--- a/spec/unit/matchers/run_spec.rb
+++ b/spec/unit/matchers/run_spec.rb
@@ -1,0 +1,96 @@
+require 'spec_helper'
+
+describe RSpec::Puppet::FunctionMatchers::Run do
+  it 'should call the lambda with no params' do
+    received_params = nil
+    if Puppet.version.to_f >= 4.0
+      subject.matches?(lambda { |env, *params| received_params = params })
+      expect(received_params).to eq([])
+    else
+      subject.matches?(lambda { |params| received_params = params })
+      expect(received_params).to be_nil
+    end
+  end
+
+  it 'should not match a lambda that raises an error' do
+    expect(subject.matches?(lambda { |env, *params| raise StandardError, 'Forced Error' })).to be_false
+  end
+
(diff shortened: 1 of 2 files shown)


DavidS:
I've added a few more improvements to the code.

spec/unit/matchers/run_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      subject.matches?(lambda { |params| received_params = params })
+      expect(received_params).to be_nil
+    end
+  end
+
+  it 'should not match a lambda that raises an error' do
+    expect(subject.matches?(lambda { |env, *params| raise StandardError, 'Forced Error' })).to be_false
+  end
+
+  [ [], [true], [false], [''], ['string'], [nil], [0], [[]], [{}], [1.1], ['one', 'two'] ].each do |supplied_params|
```
igalic: what about `:undef` ?

DavidS: good call. fixed. Also reordered the values, added a hash with a key, and added `:undef` to all other lists

DavidS:
I've added another commit to fix up the 'should' part of the failure message.

PR #288, opened 2015-05-13
Add environmentpath setting required for puppet 4

DavidS:
This setting, copied from rspec-puppet's own spec_helper.rb, allows puppet 4 to find custom functions in the test fixtures (which include symlinks to the module under test).


DavidS:
Cleaned up, thanks to @kylog.

PR #291, opened 2015-05-26
Use the correct lexer for Puppet 4 in setup.rb

alexharv074:
Without this patch applied, rspec-puppet-init errors out with
    `get_module_name_from_file': uninitialized constant Puppet::Parser::Lexer (NameError)
for Puppet 4.

### DIFF
--- a/lib/rspec-puppet/setup.rb
+++ b/lib/rspec-puppet/setup.rb
@@ -55,6 +55,8 @@ module RSpec::Puppet
     end
 
     def self.get_module_name_from_file(file)
+      # FIXME: see discussion at
+      # https://github.com/rodjek/rspec-puppet/issues/290
       if Puppet.version.to_f >= 4.0
         p = Puppet::Pops::Parser::Lexer2.new
       else


lib/rspec-puppet/setup.rb
```diff
@@ -55,7 +55,11 @@ def self.get_module_name
     end
 
     def self.get_module_name_from_file(file)
-      p = Puppet::Parser::Lexer.new
+      if Puppet.version.to_f >= 4.0
+        p = Puppet::Pops::Parser::Lexer2.new
```
igalic: @hlindberg is there any less worse way to do that?
is that interface even supposed to be used / "public" / "documented" ?

hlindberg: @igalic It is not documented and not marked as @api ... I think that is a telling sign. OTOH, the old Lexer was also not considered API. However, the entire method is flawed since it is only using the lexer - it is completely unknown if the read content is valid (e.g. "class class Foo define define class" lexes just fine :-)  The correct way is to use the parser and get the parsed result (and look for validation errors). The definitions are found inside the parsed result. With the 3x parser evaluation was sort of intertwined with parsing, and you would have to evaluate the logic to find the definitions (it scanned for them when importing/evaluating) - and thus it was very tricky to find the definitions.

Albeit also not API, the EvaluatingParser class provides high level access to parsing, validation and evaluation. I would use it - to a) parse, b) get the Program instance (program) from the parsed result, c) get the definitions from it (program.definitions), c) iterate over those to find ResourceTypeDefinition instances, and HostClassDefinition instances, and get each definition.name d) split the name and get the first part. (Note that there are other model classes that are definitions than ResourceTypeDefinition and HostClassDefinition). See the file model_meta.rb for the schema/model of everything in the AST-model.

alexharv074: I can confirm that the method is flawed, and it returns the wrong result for the actual puppet code I'm using here as input.  But, as noted, it was broken against Puppet 3 as well.  As a quick and dirty hack, though, I'm sure this hack lets lots of people quickly set up their puppet-rspec.  Perhaps I should add a "FIXME" note per @hlindberg's comments?

hlindberg: Add a ticket and reference with a FIXME in code perhaps?

alexharv074: OK, is this something that should be in Jira, or do you mean reference the issue in @rodjek's github project?

igalic: @alexharv074 if you find anything in Puppet's guts that need fixing, or at least polishing that would go there.

any issues with rspec-puppet should be filed here.

alexharv074: got it - thanks.

DavidS:
This fixes the reported problem for me, without causing regressions when using puppet 4. So that's a net win for me.

alexharv074:
@igalic OK - updated.

daenney:
Looks like this needs a rebase now.

alexharv074:
Why do you say a rebase is required @daenney - I just did a git pull on master and didn't bring in anything new?

alexharv074:
OK - looks like someone already merged the first commit?  Not sure what's going on but anyway I've rebased and pushed again which just adds the README comment.

DavidS:
I'm sorry for the confusion, that happened while preparing the 2.2 release. I should have added the FIXME myself, when merging in this PR manually.

alexharv074:
No worries!

PR #296, opened 2015-05-28
travis: enable testing against new puppet releases

DavidS:


.travis.yml
```diff
@@ -23,6 +25,8 @@ matrix:
       env: RSPEC_VERSION=2.14
     - rvm: 1.8.7
       env: PUPPET_VERSION=4.0
+    - rvm: 1.8.7
+      env: PUPPET_VERSION=4.1
```
igalic: huh?
i thought 4.x is running 2.1 only??

DavidS: It only complained about 1.8, so ¯\_(ツ)_/¯

kylog: Puppet 4.x, when bundled in puppet-agent packages does come with ruby 2.1, and that's all that we run puppet acceptance tests against, *but* we run spec tests for puppet 4 (via travis) against 1.9/2.0/2.1/2.2 and I suspect gem/source/etc installs of puppet may be running against misc rubies in that range.

So the testing bar is different, but here I don't see any downside (besides testing time) to testing against that same matrix.

igalic: oooh! i didn't know that puppet worked with ruby 2.2 already!

kylog: > oooh! i didn't know that puppet worked with ruby 2.2 already!

Just to be pedantic: I'd apply `s/worked with/is spec tested with/`  :)

Less pedantically, it *is* more lightly tested against 2.2, so there may be issues we discover as it gets used more widely in ruby 2.2 environments.

DavidS:
Merged that in #305.

PR #300, opened 2015-06-12
Break out spec_helper and rake_task

DavidS:
This way the default templates become soothingly stupid and updates to
the files will reach consumers of this code. Special requirements can
still be accommodated by not including these files.

This deliberately leaves the puppet-lint rake task in the template as
that's something people will definitely want to adjust based upon their
local requirements.


DavidS:
Merged that in #305.

PR #302, opened 2015-06-14
fix rspec-puppet documentation bug for hiera usage

hesco:
http://ask.puppetlabs.com/question/60/what-is-the-recommended-method-for-testing-modules-that-use-hiera/?answer=17865#post-id-17865 

this fix to a bug in the documentation would have saved me a couple of hours today.  

-- Hugh

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,12 @@ end
 
 ### Configuration
 
+```ruby
+# spec/spec_helper.rb 
+require 'rspec-puppet'
+require 'hiera'
+```
+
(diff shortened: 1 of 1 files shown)


DavidS:
I don't think that's how it is supposed to work. Would you mind taking a look at https://github.com/rodjek/rspec-puppet/pull/303 whether that clears up the issues?

DavidS:
Hi, I've merged a different addition to the docs for this. Please come back on this, should you have more questions.

PR #303, opened 2015-06-15
Add spec to test and demo hiera integration

DavidS:
This spec configures hiera for a test run and shows how this works with
and without a set hiera_config file.


DavidS:
This fails on 2.7, not finding the hiera function. I'll need to look into that later.

DavidS:
Merged that in #305.

PR #304, opened 2015-06-16
Add additional resources to catalog

mcanevet:
Resources declared in `generate` method can't be tested because they are added to the catalog later in the process.

This does not work yet...
I open a PR for comments.

I'm currently stuck with this when running unit tests:

```
undefined method `builddepends' for Stage[main]{:name=>"main"}:Puppet::Resource
```

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -198,20 +198,23 @@ module RSpec::Puppet
 
       # trying to be compatible with 2.7 as well as 2.6
       if Puppet::Resource::Catalog.respond_to? :find
-        Puppet::Resource::Catalog.find(node_obj.name, :use_node => node_obj)
+        catalog = Puppet::Resource::Catalog.find(node_obj.name, :use_node => node_obj)
       elsif Puppet.version.to_f >= 4.0
         env = Puppet::Node::Environment.create(
           environment,
           [File.join(Puppet[:environmentpath],'fixtures','modules')],
           File.join(Puppet[:environmentpath],'fixtures','manifests'))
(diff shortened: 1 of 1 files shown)


mcanevet:
@DavidS any idea?

mcanevet:
The relationship_graph may not be initialized...
It's maybe the same problem we have here : https://github.com/rodjek/rspec-puppet/pull/261

DavidS:
Currently in a hurry, but I think this is explained in https://github.com/puppetlabs/puppet/blob/master/docs/catalogs.md : there are two kinds of catalog, and you seem to be dealing here with the resource, not the RAL one, but relationships only really exist in the latter.

mcanevet:
@DavidS I tried to add `.to_ral` here : https://github.com/rodjek/rspec-puppet/blob/master/lib/rspec-puppet/support.rb#L222 (without any other modifications) and `rake spec` yields a lot...
It would be very nice to use RAL catalog instead of resources catalog so that we can use the `relationship_graph` method so that we can add generated resources to it and also to have a more clever way to test for resources dependencies (https://github.com/rodjek/rspec-puppet/pull/261) with the `relationship_graph` method.
But it looks like the `.to_ral` method does not just add some features...

DavidS:
yes, to_ral is actually called on the agent, and does a ton of things :-/ sorry, I don't have abetter answer right now ...

DavidS:
@mcanevet to_ral is now used on the catalog since the transitive dependency checking work was merged. You might want to revisit this now?

mcanevet:
I doing some cleanup in my old PRs. I'll close this one as I'm not interested in working on this anymore.

PR #314, opened 2015-07-15
Recursive dependencies

DavidS:


mcanevet:
@DavidS I think we can make a more clever thing by converting the catalog with to_ral and parse the graph.

raphink:
@DavidS I could have rebased #261, too ;-)

DavidS:
I just merged a fix to one of the tests that was polluting the spec environment, and wanted to see whether that would fix this too, but no avail.

I'd love to see this code being based off `to_ral` though. Did I hear someone volunteer? :-)

PR #317, opened 2015-07-30
Issue 295:  Allows structured facts to work in 'let :facts' block

bkeyser:
The change will

* allow structured facts tow work in 'let :facts' block as described in Issue 295
* when using symbols in the facts block, converts symbols to strings throughout rather than just the top-level key

If you would prefer, I could create an issue for the latter and submit it independently.

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -197,9 +197,9 @@ module RSpec::Puppet
 
       stub_facts! facts_val
 
-      node_obj = Puppet::Node.new(nodename)
+      node_facts = Puppet::Node::Facts.new(nodename, facts_val)
 
-      node_obj.merge(facts_val)
+      node_obj = Puppet::Node.new(nodename, { :parameters => facts_val, :facts => node_facts })
 
       # trying to be compatible with 2.7 as well as 2.6
       if Puppet::Resource::Catalog.respond_to? :find
@@ -227,10 +227,18 @@ module RSpec::Puppet
       @@cache[args] ||= self.build_catalog_without_cache(*args)
     end
 
+    # Facter currently supports lower case facts.  Bug FACT-777 has been submitted to support case sensitive
+    # facts.
     def munge_facts(facts)
-      output = {}
-      facts.keys.each { |key| output[key.to_s] = facts[key] }
-      output
+      return facts.reduce({}) do | memo, (k, v)|
+        memo.tap { |m| m[k.to_s.downcase] = munge_facts(v) }
+      end if facts.is_a? Hash
+
+      return facts.reduce([]) do |memo, v|
+        memo << munge_facts(v); memo
+      end if facts.is_a? Array
+
+      facts
     end
 
     def escape_special_chars(string)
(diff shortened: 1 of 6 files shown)


spec/classes/facts_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
+
+describe 'structured_facts' do
+  family = 'RedHat'
+
+  context 'should compile using symbols' do
+    let(:facts) {{
+      :os => {
+        :family => family
```
DavidS: please use a string literal here, like you do below in L20

spec/classes/facts_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  context 'should compile using strings' do
+    let(:facts) {{
+      'os' => {
+        'family' => 'RedHat'
+      }
+    }}
+
+    it { should contain_class('structured_facts') }
+    it { should compile.with_all_deps }
```
DavidS: Please add a test for the actual notifies here, so that the passed value in the notify title is actually checked for being correct.

something like

```
it { should contain_notify("RedHat") }
```

Also, I suspect that the notifies you create should cause a collision, because there should be two Notify[RedHat] resources?

DavidS:
This looks great, I have added just a few minor notes to the source.

Before merging, please squash this down to a single commit off the current master, so the history is nice and clean.

lib/rspec-puppet/support.rb
```diff
@@ -228,9 +228,15 @@ def build_catalog(*args)
     end
 
     def munge_facts(facts)
-      output = {}
-      facts.keys.each { |key| output[key.to_s] = facts[key] }
-      output
+      return facts.reduce({}) do | memo, (k, v)|
```
bkeyser: David:  I want to be sure that this is all right given your comment below about using a string literal rather than the symbol in the fact.  I don't have an understanding for why rspec-puppet requires the use of string literals in nested data structures--I know that facter does not support the use of symbols, but have often thought that rspec-puppet should convert symbols as a matter of convenience.

DavidS: Sorry I missed this before my week off.

When I was talking about the string literal I meant not using the `family = 'RedHat'` variable. Unless you want to iterate over multiple values, using a variable instead of replicating the value is just obscuring what's going on.

DavidS: If you want to support strings and symbols as *keys* in the hash, please add test cases for both to show that it's working. especially if it needs to be special-cased within rspec-puppet.

bkeyser: I hope you had a great week off :)

With your feedback, I aligned the munge_facts method to convert all symbols to lowercase strings to align with the current behavior of Puppet and included appropriate test cases.

DavidS:
Indeed it was a great week.

Even better coming back and finding good code :-) I love the new tests and would say the code's done now. If you could add a note to the README (e.g. around L366 describing structured facts as a possibility) and a new "Unreleased" section at the top of the Changelog extolling the virtues of your feature and yourself? ;-)

bkeyser:
Thanks David!

I added a description of the functionality in the README and in the CHANGELOG.  Hopefully the description make sense :)

DavidS:
Thank you very much! Excellent work.

Issue #215, opened 2014-07-25
Cached catalogs + hiera-puppet-helper


barrkel-m20:
I noticed, after several hours of debugging, that catalogs appear to be cached in a class variable and end up being reused across multiple Rspec examples even though parameterization data (`let(:hiera_data) {}` via https://github.com/mthibaut/hiera-puppet-helper) has changed.

This line does the caching:

https://github.com/rodjek/rspec-puppet/blob/389f99ef666521fec1b4530fe69dc1ab84a060a8/lib/rspec-puppet/support.rb#L162

Why is this? It doesn't appear to be responsible for a major speedup, as far as I can tell, yet has major potential for test pollution.

The symptom is that later run examples get hiera data from earlier examples.

shulima:
I was trying to figure out why enabling hiera lookup through shared context with metadata wasn't working as intended. Turned out that depending on test order, catalog cached from the hiera-enabled test was polluting the tests that were specifically supposed not to see hiera. Thank you for the extremely helpful workaround.

jhoblitt:
Kudos to @ngiger !

pall-valmundsson:
I came across this issue not because catalog reuse but because of memory consumption of the cache. 
The above modifications have performance issues when running many examples against the same catalog as the catalog has to be recompiled for each example (it's noticeable for large catalogs). My version tries to re-use the cache for serially run test cases but purges the cache as soon as new arguments are sent to build_catalog.

This is not heavily tested :)

```
module RSpec::Puppet
  module Support
    def build_catalog(*args)
      if @@cache.has_key?(args)
        @@cache[args]
      else
        @@cache = {}
        @@cache[args] ||= self.build_catalog_without_cache(*args)
      end
    end
  end
end
```

domcleal:
@pall-valmundsson I've hopefully addressed the memory issue in #329 by limiting the number of cached catalogues, while still keeping multiple in memory.

The Hiera issue here's still valid I think, the only invalidation at the moment is based on :hiera_config, not :hiera_data.

PR #329, opened 2015-10-22
Catalogue cache improvements

domcleal:
Fixes two issues in the catalogue cache:

1) Clones the facts hash to stop the `_timestamp` fact (dynamically generated by Puppet 3.x) from being part of the cache key.  This was preventing the catalogue cache from having any effect on adjacent examples on 3.x.  [Puppet 4 no longer adds this fact](https://tickets.puppetlabs.com/browse/PUP-3130).

2) Limits the size of the cache, preventing memory exhaustion when many different sets of facts and parameters were tested.

Originally reported at https://github.com/rodjek/rspec-puppet/issues/215#issuecomment-145708335.

I'm not sure about the choice of 16 catalogues.  I was tempted to only cache a single catalogue, which would allow adjacent examples (with identical parameters) to get the benefit of caching, but I suppose this might allow for some caching in more complex specs with non-linear layouts of examples (e.g. testing parameter A, then parameter B, then parameter A again).

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -2,6 +2,7 @@ module RSpec::Puppet
   module Support
 
     @@cache = {}
+    @@cache_lra = []
 
     def subject
       lambda { catalogue }
@@ -224,7 +225,19 @@ module RSpec::Puppet
     end
 
     def build_catalog(*args)
-      @@cache[args] ||= self.build_catalog_without_cache(*args)
+      unless @@cache.has_key? args
+        @@cache[args] = self.build_catalog_without_cache(*args)
+        @@cache_lra << args
+
+        # Keep only the most recently added 16 entries to prevent high memory consumption
+        expire_cache(@@cache, @@cache_lra, 16)
+      end
+      @@cache[args]
+    end
+
+    def expire_cache(cache, lra, max)
+      expired = lra.slice!(0, lra.size - max)
+      expired.each { |key| cache.delete(key) } if expired
     end
 
     # Facter currently supports lower case facts.  Bug FACT-777 has been submitted to support case sensitive
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -197,7 +197,7 @@ module RSpec::Puppet
 
       stub_facts! facts_val
 
-      node_facts = Puppet::Node::Facts.new(nodename, facts_val)
+      node_facts = Puppet::Node::Facts.new(nodename, facts_val.dup)
 
       node_obj = Puppet::Node.new(nodename, { :parameters => facts_val, :facts => node_facts })
 
--- a/spec/classes/catalogue_cache_spec.rb
+++ b/spec/classes/catalogue_cache_spec.rb
@@ -1,0 +1,43 @@
+require 'spec_helper'
+
+describe 'test::bare_class' do
+  describe 'cache between adjacent examples' do
+    catalogue_id = nil
+
+    it 'records the initial catalogue ID' do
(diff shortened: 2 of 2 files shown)


spec/classes/catalogue_cache_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+          { 'iteration' => i }
+        end
+
+        it 'should receive a new catalogue ID' do
+          expect(catalogue.object_id).not_to eq(catalogue_ids[i])
+        end
+      end
+    end
+
+    (16..20).each do |i|
```
DavidS: shouldn't that be (5..20) to cover all bases?

lib/rspec-puppet/support.rb
```diff
@@ -224,7 +225,18 @@ def stub_facts!(facts)
     end
 
     def build_catalog(*args)
-      @@cache[args] ||= self.build_catalog_without_cache(*args)
+      unless @@cache.has_key? args
+        # Keep only the most recently added 16 entries to prevent high memory consumption
+        expire_cache(@@cache, @@cache_lra, 15)
```
DavidS: It makes me sad that the API here requires the caller to do math in their head.

domcleal: Ah, the off-by-one error?

spec/classes/catalogue_cache_spec.rb
domcleal: It should be (9..20) or even (9..20) + (1..4), as the 1..20 will have invalidated the first four, then 1..4 will have invalidated the next four.  What's left should be the original 9..20 (since the first eight are now invalid), plus 1..4.

lib/rspec-puppet/support.rb
DavidS: Yes.

PS: "The two big problems in computer science: Cache Invalidation, Naming, and Off-by-one Errors."

domcleal: I've moved the line down so that it's below the new addition to the cache, meaning `16` can be passed into `expire_cache`.  Since it's the most recently added, there's no danger of it being immediately expired again.

spec/classes/catalogue_cache_spec.rb
DavidS: Right.

lib/rspec-puppet/support.rb
domcleal: Heh, two of three in one PR's not bad going!

DavidS:
I like the code change, and I'll try it out tonight. Then merge it.

domcleal:
Thanks for reviewing David.  My own unscientific test on Travis CI for one of our modules just completed with interesting results - most runs are significantly quicker and a couple of runs that I think were being OOM killed are now passing.

Before: https://travis-ci.org/theforeman/puppet-foreman/builds/85269582
After: https://travis-ci.org/domcleal/puppet-foreman/builds/86802997

lib/rspec-puppet/support.rb
igalic: we *could* bikeshed the name :P

igalic:
@domcleal why on earth are you tests running for 5 hours?!

domcleal:
@igalic I think that's totalling the runtime from all parts of the matrix.  They're quite large test suites which use [rspec-puppet-facts](https://github.com/mcanevet/rspec-puppet-facts/) to run examples on each supported OS.  The wall time was closer to 1-2 hours.

DavidS:
This improved the local runtime of the puppetlabs-apache spec tests from 11 minutes 32 seconds to slightly over 8 minutes.

Excellent catch, @domcleal , thanks!

PR #333, opened 2015-11-28
Remove catalogue cache instance variables after each example

domcleal:
Example groups provided by rspec-puppet use caching within instance
variables for catalogues, compilers etc., but these result in a memory
leak when large numbers of example groups and large catalogues are used.

rspec keeps the example group instance in memory until the end of the
suite and instance variables will simply remain too.  Clearing them
allows unused catalogues to be GCed.

theforeman/puppet was seeing process sizes of at least 3-4GB before
OOMing under MRI Ruby 2.0.0 and 2,416 examples.  This change allows it
to execute the suite with a max resident size of under 1GB.

---

#329 didn't fully fix our OOM issues.  I think it probably reduced the number of catalogues generated and fixed some tests, but left others.

Before this PR: https://travis-ci.org/theforeman/puppet-puppet/builds/93412213 ("Killed"/OOMs on Ruby 2.0.0)
With this PR: https://travis-ci.org/domcleal/puppet-puppet/builds/93687348

### DIFF
--- a/lib/rspec-puppet/example.rb
+++ b/lib/rspec-puppet/example.rb
@@ -40,4 +40,7 @@ RSpec::configure do |c|
     c.include RSpec::Puppet::ProviderExampleGroup, :type => :provider, :file_path => c.escaped_path(%w[spec providers])
   end
 
+  # Hook for each example group type to remove any caches or instance variables, since they will remain
+  # and cause a memory leak.  Can't be assigned per type by :file_path, so check for its presence.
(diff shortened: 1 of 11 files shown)


roidelapluie:
@domcleal there is an incompatibility with rspec 2.14

domcleal:
@roidelapluie yeah, I fixed that separately in #334 (it's actually with Puppet 4.3, which that test's using).

roidelapluie:
I just saw #334 indeed. thanks.

roidelapluie:
:+1: for the merge

domcleal:
Rebased after the merge of #334.

PR #334, opened 2015-11-28
Fix expected argument mismatch message on Puppet 4.3.0

domcleal:
The exception message raised under Puppet 4.3.0 for arity mismatches in
functions changed:

    expected split("foo") to have raised ArgumentError matching /mis-matched arguments/
    instead of ArgumentError('split' expects 2 arguments, got 1)


roidelapluie:
:+1:

PR #335, opened 2015-11-29
Allow undef parameter passing with symbols

### LINKED ISSUE #279: Can't pass undef as a param value

domcleal:
Now, when a symbol is used as a param value, its `inspect` method is not
called to construct the Puppet parameter value. Instead, the unquoted
string, e.g. `undef` is used in the generated resource declaration. So now
`let(:params) { { :foo => :undef } }` leads to the desired resource
declaration `class { 'test': foo => undef }`.

Before this change, it was impossible to pass `undef` as a parameter.
Using `nil`, `:undef`, or `'undef'` all failed for various reasons.

Credit to the discussion at [1] for identifying the issue and proposing
an alternate solution.

Fixes #279.

[1] <https://groups.google.com/forum/#!topic/puppet-users/6nL2eROH8is>

Adapted by Dominic Cleal from https://github.com/rodjek/rspec-puppet/pull/280 to use symbols instead of nil.

---

Replaces #280.  The idea of using a class is OK as a workaround, but unnecessary I think for proper support in rspec-puppet.  Using a symbol - which has no function in r-p today, as #inspect generates an invalid Puppet DSL string - seems lightweight and may have other uses in future.

I can imagine using it to pass more complex verbatim strings in, e.g. `let(:params) { {:param => :'file(..)'} }` or similar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,12 @@ let(:title) { 'foo' }
 let(:params) { {:ensure => 'present', ...} }
 ```
 
+##### Passing Puppet's `undef` as a parameter value
+
+```ruby
+let(:params) { {:user => :undef, ...} }
+```
+
 #### Specifying the FQDN of the test node
 
 If the manifest you're testing expects to run on host with a particular name,
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -122,8 +122,13 @@ module RSpec::Puppet
 
     def param_str
       params.keys.map do |r|
-        param_val = escape_special_chars(params[r].inspect)
-        "#{r.to_s} => #{param_val}"
+        param_val = params[r]
+        param_val_str = if param_val == :undef
+                          'undef'  # verbatim undef keyword
+                        else
+                          escape_special_chars(param_val.inspect)
+                        end
+        "#{r.to_s} => #{param_val_str}"
       end.join(', ')
     end
 
--- a/spec/classes/undef_spec.rb
+++ b/spec/classes/undef_spec.rb
@@ -1,0 +1,18 @@
+require 'spec_helper'
+
+describe 'undef_test' do
+  it { should compile.with_all_deps }
+
+  shared_examples 'exec echo' do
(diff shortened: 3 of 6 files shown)


DavidS:
I'm not very convinced that using symbols beyond `:undef` is a good idea, given that they'll no be GC'd. having an extensions to the DSL, like `pp('file(..)')` that passes the string verbatim into the manifest might be cleaner.

As a first step I would prefer this change to only accept `:undef` as a value, so we don't run into issues later.

domcleal:
Good points, you're right and I like the idea of a `pp` method or similar.   I've updated the PR to support only :undef.

DavidS:
Thanks a lot!

PR #337, opened 2015-12-09
Extract cache class from support mixin

adrienthebo:
The RSpec::Puppet::Support mixin has become a god object and has gobs of
unrelated functionality, making it hard to extend/refactor/improve. This
commit starts simplifying it by extracting a real cache object to a
class and using that instead of the inline cache.


domcleal:
:+1: nice refactoring, thanks!

PR #338, opened 2015-12-10
Respect rspec module path manifests

adrienthebo:
When @binford2k was trying to use rspec-puppet to test entire environments with Puppet 4.x, he found out that rspec-puppet hardcodes the module_path and manifest settings and thus can only use the standard fixture paths for module testing. This commit attempts to fix this by testing to if the RSpec configuration settings are set and preferring those, then falling back to the previously set fixture paths.

/cc @binford2k; I would like your help testing this.


DavidS:
Looks good to me.

danzilio:
:+1:

adrienthebo:
Unfortunately this doesn't correct the hard coded paths in coverage.rb; should we pass the coverage object the environment paths with a before all hook?

In addition the module_path rspec setting has a default value; it should only apply to puppet 3.x but it might be possible for that value to gum things up.

binford2k:
:+1: totally works for me

DavidS:
@adrienthebo I'd like to cut a release before the holidays. Do you feel the code is already releasable and can be merged?

On the coverage point: I'd say, using `RSpec.configuration.module_path || ...` there the same way as in `build_4x_environment` would be an adequate workaround for now. Long-term I'd love to see a simpler and more flexible way to influence puppet's config.

On the point of the default setting, I guess that all tests that need it already override this, because the default (`/etc/puppet/modules`) is not a good place to use during tests. That said, this value should probably go away.

adrienthebo:
I can drop the default `module_path` configuration value; that'll clear up that class of problems. The tricky thing with the Coverage object is that there's not a well defined lifecycle for the object; this makes it tricky to supply reasonable values to the object. I'll do some tinkering around this and see if I can get something reasonable worked out.

adrienthebo:
I believe that I've fixed the coverage issues. We should try to get this PR merged for the upcoming release; once that release has shipped I would like to do a more in depth refactoring of this code.

PR #340, opened 2015-12-15
Refactor coverage.rb

adrienthebo:
This pull request refactors coverage.rb and replaces the use of the `Singleton` mixin with a simpler singleton implementation that allows for instances to be created and tested. It also replaces the use of the `at_exit` Ruby hook with the RSpec `after(:suite)` hook since we're already working within the RSpec ecosystem.


binford2k:
:+1:

PR #341, opened 2015-12-15
Extract version specific Puppet functionality

adrienthebo:
The rspec-puppet codebase is littered with code that changes behavior based on what version of Puppet is available; this makes the code harder to understand and reason about, increases the cost of maintenance, and frequently leads to code duplication. This pull request is an initial attempt to pull version specific behavior into a series of classes that can share common behavior and override version specific behavior at a single point.

### DIFF
--- a/lib/rspec-puppet/adapters.rb
+++ b/lib/rspec-puppet/adapters.rb
@@ -26,6 +26,10 @@ module RSpec::Puppet
       def catalog(node, _)
         Puppet::Resource::Catalog.indirection.find(node.name, :use_node => node)
       end
+
+      def environment(name)
+        Puppet::Node::Environment.new(name)
+      end
     end
 
     class Adapter4X < Base
@@ -40,8 +44,8 @@ module RSpec::Puppet
         ]
       end
 
-      def catalog(node, environment)
-        env = build_4x_environment(environment)
+      def catalog(node, environment_name)
+        env = environment(environment_name)
         loader = Puppet::Environments::Static.new(env)
         Puppet.override({:environments => loader}, 'Setup test environment') do
           node.environment = env
@@ -49,9 +53,7 @@ module RSpec::Puppet
         end
       end
 
-      private
-
-      def build_4x_environment(name)
+      def environment(name)
         modulepath = RSpec.configuration.module_path || File.join(Puppet[:environmentpath], 'fixtures', 'modules')
         manifest = RSpec.configuration.manifest || File.join(Puppet[:environmentpath], 'fixtures', 'manifests')
         Puppet::Node::Environment.create(name, [modulepath], manifest)
--- a/lib/rspec-puppet/adapters.rb
+++ b/lib/rspec-puppet/adapters.rb
@@ -30,18 +30,23 @@ module RSpec::Puppet
       def environment(name)
         Puppet::Node::Environment.new(name)
       end
+
+      def settings_map
+        [
+          [:modulepath, :module_path],
+          [:config, :config],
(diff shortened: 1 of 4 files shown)


binford2k:
:+1:

igalic:
wow… that reads so much nicer :100:

lib/rspec-puppet/adapters.rb
```diff
@@ -0,0 +1,107 @@
...
+      def set_setting(example_group, puppet_setting, rspec_setting)
+        if example_group.respond_to?(rspec_setting)
+          value = example_group.send(rspec_setting)
+        else
+          value = RSpec.configuration.send(rspec_setting)
+        end
+        begin
+          Puppet[puppet_setting] = value
+        rescue ArgumentError
+          # TODO: this silently swallows errors when applying settings that the current
```
DavidS: Could this at least emit a message that a setting is swallowed? Having the message should go a long way towards mitigating surprises.

DavidS:
A minor comment on swallowing the ArgumentException, but great otherwise! I'm looking forward to moving more stuff into those adapters.

lib/rspec-puppet/adapters.rb
adrienthebo: Kinda. This could cause a large number of warnings to be emitted; I was just (mostly) preserving the existing behavior so this isn't a regression. Should we fix this here or should we spin this out into another issue?

DavidS: It's not worth holding this up. If and when this is revisited, we might want to add a depreciation warning using https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/warnings.rb ; those are only printed once, and supply infrastructure to trace where they originated.

adrienthebo: @DavidS I just looked into those warnings and all of those methods/modules are marked private: https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/warnings.rb#L5

DavidS: Ah, the real interface seems to be here: https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/warnings.rb

adrienthebo: Aaaactually... those are also private :/ https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/warnings.rb#L19-L21

From the rspec-support readme:

```
RSpec::Support provides common functionality to RSpec::Core, RSpec::Expectations and RSpec::Mocks. It is considered suitable for internal use only at this time.
```

:/

However, the implementation of warnings is really simple; it's just calling `Kernel.warn` by default (https://github.com/rspec/rspec-support/blob/master/lib/rspec/support.rb#L108-L119). At this point I think it would be less trouble for us to implement whatever warning system we needed.

I'm going to spin out a new issue for this.

DavidS: Thanks for the great writeup, although https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/warnings.rb#L7 doesn't seem to be private. The two main features around warning I'm advocating for is supressing multiple logs of the same warn and the ability to show a stack trace for each warn.

adrienthebo: It's not the file itself that's private, it's actually the entire project: https://github.com/rspec/rspec-support/blob/master/README.md#rspecsupport . It's also a new dependency; apparently rspec-core doesn't have a hard dependency on it - or at least the version I have installed.

Anywho. Happy to use that if you like, but let's bounce this conversation over to the ticket :)

Issue #343, opened 2015-12-29
Puppet versions >4 do not support full modulepath


dylanratcliffe:
Logic [here](https://github.com/rodjek/rspec-puppet/blob/master/lib/rspec-puppet/coverage.rb#L129) shows that for some reason we are not splitting the modulepath on a ':' if we are using a Puppet version greater than 4. Why are we doing this? I'm getting broken tests now as I cannot set more than one path as the modulepath.

My tests using older version: [2.2.0](https://travis-ci.org/dylanratcliffe/puppet_controlrepo/builds/92623572)
My tests using current version: [2.3.0](https://travis-ci.org/dylanratcliffe/puppet_controlrepo/builds/99192834)

DavidS:
That slipped through the cracks when refactoring this section in https://github.com/rodjek/rspec-puppet/commit/e44da91788284fe61939adce006b6b3704725f21 . I guess it doesn't hurt splitting  RSpec.configuration.module_path on colons.

I'm tempted to add that puppet 4 has deprecated the global modulepath setting (https://docs.puppetlabs.com/references/latest/configuration.html#modulepath) , but I realize that it's not relevant to this bug.

adrienthebo:
See #349, @dylanratcliffe could you test that pull request?

dylanratcliffe:
@adrienthebo Sorry about the delay, been out training. Seems like more has broken now, [here are the results of my tests](https://travis-ci.org/dylanratcliffe/puppet_controlrepo/builds/103006059). And the [actual error](https://travis-ci.org/dylanratcliffe/puppet_controlrepo/jobs/103006062#L376). 

Here is what the spec_helper.rb looks like:
```ruby
require 'puppetlabs_spec_helper/module_spec_helper'

RSpec.configure do |c|
  c.parser = 'future'
  c.module_path = '/var/folders/tv/svz3wzy550j8dktpxs_vqprm0000gn/T/r10k20160118-54626-g4wxeq/etc/puppetlabs/code/environments/production/site:/var/folders/tv/svz3wzy550j8dktpxs_vqprm0000gn/T/r10k20160118-54626-g4wxeq/etc/puppetlabs/code/environments/production/modules'
  c.hiera_config = '/var/folders/tv/svz3wzy550j8dktpxs_vqprm0000gn/T/r10k20160118-54626-g4wxeq/etc/puppetlabs/code/environments/production/hiera.yaml'
end
```

The test it is running is just an `it { should compile }` test

[...]

PR #349, opened 2016-01-09
GH-343: Support colon separated modulepaths and environmentpaths

adrienthebo:
The addition of Puppet 4 support added a number of separate code paths for Puppet 4. This caused a number of instances to creep in where colon delimited modulepaths were supported, but not on Puppet 4. Moreover Puppet 4 calculates the current modulepath and manifest relative to the current environment, which is quite different from how Puppet 3 and earlier traditionally handled it.

This pull request introduces a number of changes in order to consistently handle modulepaths and manifests across multiple versions of puppet

  * Eagerly initialize Puppet setttings so that the current environment can be set before specs are run
  * Always use the current environment when getting catalogs and functions 
  * Properly set up the Puppet 4.x context with correct environment information
  * Properly split the modulepath on Puppet 3.x
  * Properly split the environmentpath on Puppet 4.x

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.gem
 /Gemfile.lock
+/Gemfile.local
 /.bundle/
 /vendor/
 .ruby-version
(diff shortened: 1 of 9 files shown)


dylanratcliffe:
Works for me

Issue #353, opened 2016-01-22
Test failures with 2.3 and new Puppet 3.8.5


garethr:
I hit an issue this morning with the Docker module in Travis where a single test (which had previously passed) failed, but only under the newly released Puppet 3.8.5 (released January 21st). This was also reported independently by another contributor https://github.com/garethr/garethr-docker/pull/396.

Running the test on it's own see's it pass. Testing with previously passing code fails. Testing with rspec-puppet master and with 2.3 also fails. As suggested by @DavidS I tried 2.2 and this allowed the test to pass.

After much spelunking I was able to get a small test case which reliably recreates the failure.

```puppet
require 'spec_helper'

describe 'docker::run', :type => :define do
  let(:title) { 'sample' }
  context 'passing the required params' do
    let(:facts) { {:osfamily => 'Archlinux'} }
    let(:params) { {'command' => 'command', 'image' => 'base'} }
    initscript = '/etc/systemd/system/docker-sample.service'
    # The first 16 tests will pass. The rest will fail on Puppet 3.8.5 with
    # NoMethodError:
    # undefined method `resource' for nil:NilClass
    # /Users/garethr/.rvm/gems/ruby-2.1.5/bundler/gems/rspec-puppet-7ab588ecdd1e/lib/rspec-puppet/matchers/create_generic.rb:83:in `matches?'
    18.times do
      it { should contain_file(initscript).with_content(/docker run/).with_content(/command/) }
    end
  end
end
```

Note the magic number 16. Given that is the number of catalogues cached in rspec-puppet I think that's likely related. Whether this points to a bug in Puppet 3.8.5 which rspec-puppet is exposing, or whether it's rspec-puppet using a private API that's changed it's behaviour, or something else I'm unsure.

cardil:
This was also my issue. I've created already a ticket on Puppetlabs Jira on this: https://tickets.puppetlabs.com/browse/PUP-5743

daenney:
As posted by @hlindberg in that ticket:

> With a bit more detective work. The recently introduced cache in rspec_puppet does the wrong thing when it evicts entries from its cache. (It ends up removing the entries that should be kept as opposed to those that should be evicted.
> https://github.com/rodjek/rspec-puppet/blob/8a90608341077197323c6e35c042f7d1699a3cae/lib/rspec-puppet/cache.rb#L26

hlindberg:
[...]

PR #354, opened 2016-01-22
(PUP-5743) Fix issue with expiring cache as described in #353

garethr:


garethr:
This could ideally do with a test, But it's late Friday and this appears to resolve the problem from manual testing on the Docker module.

Lots of context in https://github.com/rodjek/rspec-puppet/issues/353 and https://tickets.puppetlabs.com/browse/PUP-5743

adrienthebo:
If we've verified that this solves the issue and are confident in it then I can write tests for this later and we can ship it now.

garethr:
@adrienthebo @DavidS is just checking with the Apache module as well to get some confidence.

lib/rspec-puppet/cache.rb
```diff
@@ -23,7 +23,7 @@ def get(*args, &blk)
     private
 
     def expire!
-      expired = @lra.slice!(0, @lra.size - MAX_ENTRIES)
+      expired = @lra.slice!(0, @lra.size - (MAX_ENTRIES-1))
```
hlindberg: The -1 only helps with the case where the same args are used twice in a row at the border of MAX_ENTRIES, it does not help with the case when @ira[0] should be retained because the about to be added args are the same as @ira[0].
If args are give to expire! it could remove all those entries from @irl. In the case when the same args appear repeatedly all entries in the @ira will be the same, and there is nothing that should be evicted !

irl:
Hi, I would be happy to have those entries removed from myself. LGTM.

lib/rspec-puppet/cache.rb
```diff
@@ -11,10 +11,10 @@ def initialize(&default_proc)
     end
 
     def get(*args, &blk)
+      expire!
       if !@cache.has_key? args
         @cache[args] = (blk || @default_proc).call(*args)
         @lra << args
-        expire!
```
hlindberg: if `expire!` is kept here, this cache can be a LRU cache by doing this on the line above:
```
@irl = (@irl << 1).reverse.uniq.reverse
```
As an example given [1,2,3,4] and appending 2, the result is [1,3,4,2]. A long as cache is a small number of entries that solution is good enough. Then, when expires removes entries, it will remove the MAX_ENTRIES least recently used entries

lib/rspec-puppet/cache.rb
hlindberg: With the fix I propose above, the -1 should not be there

lib/rspec-puppet/cache.rb
cardil: That is some serious caching stuff. To deep for me to follow on Friday evening 😉

DavidS:
This change fails on puppet 3.8.5; I've went back and checked: it fails there in the same way as without this change.

running git bisect on puppet 3.8.5 points at https://github.com/puppetlabs/puppet/commit/8136aff14d36eedf99df1731cb1997c3db288039 as the cause for this issue, which plausibly could have impact on the catalogue_id or the args used as key in the cache.

Investigation continuing ...

hlindberg:
Looking at the code the bisect work found.

hlindberg:
The change in 3.8.5 is also in 4.3.2. Before the commit in question, the actual environment and the string that is used as the name of the environment for a node could diverge. That is a good thing I think.

hlindberg:
I have a theory that the changed cache logic in this PR will cause a recompile when one is not needed, it will place the new entry in the cache where the old logic would not. That would for sure cause identities to be different. However, the same failure is not seen when puppet < the commit in question, so there must be something else that causes rspec_puppet expectancy to not be true after the change. Need to dig deeper into rspec_puppet.

lib/rspec-puppet/cache.rb
hlindberg: [1,2,3,4], becomes [1,2,3,4,1] when appending 1. if you do a unique, the result is [1,2,3,4]. if instead reversing the array first it becomes [1,4,3,2,1] and a unique then gives [1,4,3,2], which reversed (again) is [2,3,4,1]. The 1 was "touched" and moved last in the array. Now if cache max was 4, and we add an element 5, we first get [2,3,4,1,5], and the expire will evict the first 3 and we end up with [1, 5]. The '1' survived because it was used later in time than 2, 3, and 4.

That is the idea at least.

There are other ways of doing the same thing - in pseudo code:
if cache-hit
  find the corresponding entry in @ira and move it last in @ira (i.e. this indicates "used last")
```
if cache-miss
   evict from cache if there are too many entries, but only evict one (seems bad to evict all of them)
  add the args to @ira, and @cache
end
```

However that is not the problem here (unless a very odd combination of args/cache is interacting). The above is however a better cache implementation if the idea is to just cache catalogs given its args (input); hard to tell if there are other requirements.

Will continue looking at the other problem.

hlindberg:
A posted #355 with what I hope is a better cache algorithm. Will be interesting to see if that changes the other problem. (Waiting for travis run there).

DavidS:
Closing this as it didn't address the stated issue.

PR #355, opened 2016-01-22
Make caching evict least recently used

hlindberg:
Before this, when the cache was full, it evicted max-cache-size number of entries from the list of entries.
It would also do eviction after addition.

With a cache size of 3, and using the sequence A, B, C, D, The old version would evict A, B, C. If again B or C was requested, then they would need to be recreated. The old would then have state D. The new will evict only A when it gets D. When it looks up B, it will also move B last, so that the state the is CDB. Next addition E, will the evict C, and state is DBE. If B is again requested, it is moved last and state s DEB.- i.e. the cache caches Most Recently Used.

I am making this change to improve the caching behavior, but also to see if it has effect on the tests that fail on catalog id being different.

### DIFF
--- a/lib/rspec-puppet/cache.rb
+++ b/lib/rspec-puppet/cache.rb
@@ -14,7 +14,7 @@ module RSpec::Puppet
       if @cache.has_key? args
         # Cache hit
         # move that entry last to make it "most recenty used"
-        @ira.insert(-1, @ira.delete_at(@ira.index(args)))
+        @lra.insert(-1, @lra.delete_at(@lra.index(args)))
       else
         # Cache miss
         # Ensure room by evicting least recently used if no space left
--- a/lib/rspec-puppet/cache.rb
+++ b/lib/rspec-puppet/cache.rb
@@ -11,10 +11,16 @@ module RSpec::Puppet
     end
 
     def get(*args, &blk)
-      if !@cache.has_key? args
+      if @cache.has_key? args
+        # Cache hit
+        # move that entry last to make it "most recenty used"
+        @ira.insert(-1, @ira.delete_at(@ira.index(args)))
+      else
+        # Cache miss
+        # Ensure room by evicting least recently used if no space left
+        expire!
         @cache[args] = (blk || @default_proc).call(*args)
         @lra << args
-        expire!
       end
 
       @cache[args]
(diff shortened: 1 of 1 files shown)


lib/rspec-puppet/cache.rb
```diff
@@ -11,10 +11,16 @@ def initialize(&default_proc)
     end
 
     def get(*args, &blk)
-      if !@cache.has_key? args
+      if @cache.has_key? args
+        # Cache hit
+        # move that entry last to make it "most recenty used"
+        @ira.insert(-1, @ira.delete_at(@ira.index(args)))
```
kylog: Should this be `lra` not `ira`?

cardil:
Linking to bug #353. Maybe this will fix it. :confused:

lib/rspec-puppet/cache.rb
hlindberg: doh, the ira is not involved in this problem afaik. Thanks

kylog:
With the `lra` fix applied, this still fails spec tests. I think this change makes sense standalone but doesn't address #353.

lib/rspec-puppet/cache.rb
```diff
@@ -23,8 +29,8 @@ def get(*args, &blk)
     private
 
     def expire!
-      expired = @lra.slice!(0, @lra.size - MAX_ENTRIES)
-      expired.each { |key| @cache.delete(key) } if expired
+      # delete one entry (the oldest) when there is no room in cache
+      @cache.delete(@lra.shift) if @cached.size == MAX_ENTRIES
```
kylog: Also this should be `@cache.size` not `@cached.size` (drop the `d`).

hlindberg: mmm, my typing skills are deteriorating - trying to make a quick fix... ha...

DavidS:
If you could rebase and squash that on top of current master, we can get a proper reading off travisci.

hlindberg:
There are quite a few cache unit tests that need to change as a consequence of this change as they test that the behavior is to drop all entries but the last when cache limit is hit. Is that actually wanted, or just someone designing tests to fit the implementation :-) ??

DavidS:
I wouldn't have thought you needed an explanation how all of this works :-p

hlindberg:
I don't have time to fix all the tests that this breaks (the tests are mostly meaningless though as they test that the cache behaves strangely). If someone wants to pick this up I am happy to help.

rodjek:
@hlindberg Apologies for the extremely delayed response! Would you mind checking the "Allow collaborators to push changes" box on the right side of this page so that I can rebase this off master?

PR #358, opened 2016-01-25
(PUP-5743) adapt to new environment name semantics

DavidS:
Since applying a fix for PUP-5522 (puppet 3.8.5 and 4.3.2) puppet symbolizes environment names
internally. The catalog cache needs to assume that the facts and other args do not change between
runs, so we have to mirror this here. Puppet versions before the change require a string as environment
name, or they fail with "Unsupported data type: 'Symbol' on node xyz"
See https://github.com/rodjek/rspec-puppet/pull/354 and PUP-5743 for discussion of this

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -9,7 +9,18 @@ module RSpec::Puppet
     end
 
     def environment
-      'rp_env'
+      # unfreeze PUPPETVERSION because of https://github.com/bundler/bundler/issues/3187
+      ver = Gem::Version.new("#{Puppet::PUPPETVERSION}")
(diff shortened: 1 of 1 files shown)


DavidS:
merged manually.

mwhahaha:
Actually with this change, i'm now getting the "Unsupported data type: 'Symbol' on node ...".  The logic seems to be backwards for the if clause. Shouldn't it be...

```
if (ver >= Gem::Version.new('3.8.5') && ver < Gem::Version.new('4.0.0')) || ver >= Gem::Version.new('4.3.2')
```

DavidS:
@mwhahaha  I just released 2.3.2 to address this issue

PR #359, opened 2016-01-25
(PUP-5743) adapt to new environment name semantics

DavidS:
Since applying a fix for PUP-5522 (puppet 3.8.5 and 4.3.2) puppet symbolizes environment names
internally. The catalog cache needs to assume that the facts and other args do not change between
runs, so we have to mirror this here. Puppet versions before the change require a string as environment
name, or they fail with "Unsupported data type: 'Symbol' on node xyz"
See https://github.com/rodjek/rspec-puppet/pull/354 and PUP-5743 for discussion of this

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -9,7 +9,18 @@ module RSpec::Puppet
     end
 
     def environment
-      'rp_env'
+      # unfreeze PUPPETVERSION because of https://github.com/bundler/bundler/issues/3187
+      ver = Gem::Version.new("#{Puppet::PUPPETVERSION}")
+      # Since applying a fix for PUP-5522 (puppet 3.8.5 and 4.3.2) puppet symbolizes environment names
+      # internally. The catalog cache needs to assume that the facts and other args do not change between
+      # runs, so we have to mirror this here. Puppet versions before the change require a string as environment
+      # name, or they fail with "Unsupported data type: 'Symbol' on node xyz"
+      # See https://github.com/rodjek/rspec-puppet/pull/354 and PUP-5743 for discussion of this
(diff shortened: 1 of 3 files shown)


lib/rspec-puppet/support.rb
```diff
@@ -9,7 +9,18 @@ def subject
...
     def environment
-      'rp_env'
+      # unfreeze PUPPETVERSION because of https://github.com/bundler/bundler/issues/3187
+      ver = Gem::Version.new("#{Puppet::PUPPETVERSION}")
+      # Since applying a fix for PUP-5522 (puppet 3.8.5 and 4.3.2) puppet symbolizes environment names
+      # internally. The catalog cache needs to assume that the facts and other args do not change between
+      # runs, so we have to mirror this here. Puppet versions before the change require a string as environment
+      # name, or they fail with "Unsupported data type: 'Symbol' on node xyz"
+      # See https://github.com/rodjek/rspec-puppet/pull/354 and PUP-5743 for discussion of this
+      if (Gem::Version.new('3.8.5') <= ver && ver < Gem::Version.new('4.0.0')) || Gem::Version.new('4.3.2') <= ver
```
cardil: This is how to introduce tech dept. I think that Puppetlabs should revert their changes, and provide a new API, rather then changing the current implementation on the fly!

DavidS: Given what we learned about the actual root cause today, a more elaborate
API would very well have had exhibited the same issue. As you can see from
the 2.3.2 release, this code didn't see the light of a second day.

Part of the process issues were fixed together with this by extending the
test matrix on rspec-puppet's side. Another part will be followed up
internally to integrate rspec-puppet's test suite into puppet's own test
setup to catch this earlier the next time around.

cardil: Yes. Good to know it turns off so slick after all.

PR #370, opened 2016-02-29
Allow testing for exported resources

### LINKED ISSUE #113: test exported resources
(this should not be mixed up with: https://github.com/rodjek/rspec-puppet/pull/51)

I would like the following functionality to be added to rspec puppet:

if a manifest has an exported resource rspec should be able to test whether the resource is exported correctly. At the moment rspec can only test for stuff that ends up in the catalogue.

nhinds:
Fixes #113 

I wasn't entirely sure where the support class which patches Puppet's internals should go, as there didn't seem to be an existing place that did anything similar, so I added a new file under `rspec-puppet/support` for this. I also tried to defend against puppet changing this class in the future by adding a couple of guards, and I changed the original code from the gist to make it more obvious that `catalogue.__rspec_puppet_exported_resources` is intended to be something internal to `rspec-puppet` and not something to rely on.

Let me know if it needs tidying up.

I've added tests for 2 matchers (`contain_X` and `have_X_resource_count`). I found that the standard `have_resource_count` matcher didn't work with this implementation, as it blindly subtracts 1 from the count (it assumes `Stage[main]` is always in the catalogue it's testing) -- I didn't want to change the matcher as part of this PR, as it didn't seem like something people would normally want to do with exported resources and I'm not sure if the `@actual_number = @actual_number - 1` line can safely be changed without breaking people's tests.


DavidS:
Looks great. If you could squash all the commits into one, that would make the git history so much nicer.

nhinds:
Squashed the commits for making it work with ruby 1.8 / rspec 2.

baloo:
Actually I think changing https://github.com/rodjek/rspec-puppet/blob/master/lib/rspec-puppet/adapters.rb#L70 by:
``` ruby
Puppet::Parser::Compiler.compile(node)
```
Should be sufficient, you don't need to monkey patch the indirector. Just don't call the indirector.
This will give you a catalog with resources like:

``` json
{
  "type": "File",
  "title": "PATH",
  "tags": [
  ],
  "file": "config.pp",
  "line": 20,
  "exported": true,
  "parameters": {
    "ensure": "present",
    "content": "BLAH",
    "mode": "0644",
    "owner": "root",
    "group": "root"
  }
}

```

baloo:
See #371
Diff: https://github.com/nhinds/rspec-puppet/compare/test-exported-resources...baloo:exported

nhinds:
I like it, looks much cleaner...

Unfortunately that has the downside of changing the existing catalogue that tests are written against to include exported resources (e.g. even without calling `exported_resources`, exported resources would match the `contains_X` matcher). That could possibly be fixed by adding `catalog.filter { |r| r.virtual? }` (which used to be called by `Puppet::Resource::Catalog::Compiler.filter`) somewhere in `rspec-puppet`...

I'm also not sure what other changes in functionality there will be due to skipping all the code in `Puppet::Resource::Catalog::Compiler.find` / `Puppet::Indirector::Indirection.find`, and whether this makes this more or less likely to cause problems due to future changes in puppet.

Maybe instead of changing the regular `RSpec::Puppet::Adapters::Base.catalog(node)` method, leave it calling the indirector and have a new method there for getting the exported resources which uses `Puppet::Parser::Compiler.compile(node)`? That way if something changes, it's only tests for exported resources which break, rather than all tests.

nhinds:
@baloo I've incorporated your change into this PR, and added a test to ensure that exported resources aren't exposed to the regular catalog.

I've tweaked it so that the indirector is called for normal catalog retrieval, and your method (calling the compiler directly) is used for exported resources, where we need to override the filter.

DavidS:
:heart: great stuff. I'll see to it that I squeeze in time to run a few more tests on this to get this merged.

baloo:
@nhinds awesome, then I'm closing #371

nhinds:
@DavidS is there anything I can do to help get this merged?

PR #371, opened 2016-03-02
Exported resources

baloo:
see #370


nhinds:
As per #370, this exposes exported resources to normal tests... addressed and incorporated into #370, but feel free to improve upon it

PR #382, opened 2016-04-07
Enable nested hashes in params

siwilkins:
Currently, it isn't possible to nest a hash within a param, so something like this:

    let(:params) do
      {
        description: 'A job with a masked password',
        masked_passwords: { 
          PASSWORD: 'abcdefg'
        }
      }
    end

fails with an error along these lines:

    
     Puppet::ParseErrorWithIssue:
       Could not parse for environment rp_env: Syntax error at ':' at line 1:117

This change enables nested hashes to be parsed correctly.

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -123,9 +123,16 @@ module RSpec::Puppet
     end
 
     def param_str
-      params.keys.map do |r|
+      param_str_from_hash(params)
+    end
+
+    def param_str_from_hash(params_hash)
+      params_hash.keys.map do |r|
         param_val = params[r]
-        param_val_str = if param_val == :undef
+        param_val_str = case param_val 
+                        when Hash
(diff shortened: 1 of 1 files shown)


siwilkins:
I've realised this breaks the param name when it's a hash (adding "[]"). I'll fix and re-submit.

DavidS:
This is awesome. Please do resubmit!

PS: if you use `push --force`, you can update a PR with new code.

siwilkins:
Ah, thanks.

As it turns out, it looks like the problem with the param name was me abusing the resulting hash in an ERB template (attempting to treat it as a normal Ruby hash).

I have spotted a different bug with my change; unfortunately I forgot to re-open this before pushing, and having now rebased it in an attempt to try again it's now not letting me re-open it. So I'll have to do a separate PR...

siwilkins:
For the record, the ERB template was actually a result of me using caps for a hash key, which it turns out causes untold misery.

PR #383, opened 2016-04-07
Enable nested hashes in params

siwilkins:
Currently, it isn't possible to nest a hash within a param, so something like this:

    let(:params) do
      {
        description: 'A job with a masked password',
        masked_passwords: { 
          db_password: 'abcdefg'
        }
      }
    end

fails with an error along these lines:

    Puppet::ParseErrorWithIssue:
      Could not parse for environment rp_env: Syntax error at ':' at line 1:117
This change enables nested hashes to be parsed correctly.

(See https://github.com/rodjek/rspec-puppet/pull/382 for slightly more background)

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -123,9 +123,16 @@ module RSpec::Puppet
     end
 
     def param_str
-      params.keys.map do |r|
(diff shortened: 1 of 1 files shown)


DavidS:
Rebased and merged with #409

PR #386, opened 2016-04-13
Fix headings under kramdown markdown processor

domcleal:
Some headings weren't rendered correctly under kramdown due to missing
whitespace, ensuring a line after each heading renders it in the HTML.

The Gemfile.lock was removed and _config.yml refreshed to match the
GitHub Pages change to Jekyll 3.0 and kramdown.


DavidS:
thanks!

Issue #387, opened 2016-04-22
Odd behavior on User type


howdoicomputer:
```
  1) appserver should compile into a catalogue without dependency cycles
     Failure/Error: it { should compile }
       error during compilation: Parameter managehome failed on User[jetty]: Validate method failed for class managehome: undefined method `manages_homedir?' for NilClass:Class
     # ./spec/classes/appserver_spec.rb:28:in `block (2 levels) in <top (required)>'
```

I'm on Puppet 2.7 and trying to move a piece of a massive codebase over to RSpec testing but I'm stuck on the portion of my module that handles manifests. I'm on a Mac and I can't update to later versions of Puppet.

I've already set my operating system values so that it chooses a provider that supports `managehome`.

```
      :operatingsystem => 'RedHat',
      :operatingsystemmajrelease => '6',
      :osfamily => 'RedHat',
      :kernel => 'Linux'
```

I tracked down the portion of code that is throwing the error:

https://github.com/puppetlabs/puppet/blob/2.7.x/lib/puppet/type/user.rb#L306

I'm a little confused here since Puppet is huge and I'm pretty inexperienced with it. I'm trying to print debug the user type to figure out what is wrong. If I don't provide an `operatingsystem` fact then I can pry out this output:

```
The provider is User[jetty](provider=directoryservice)
the class is Puppet::Type::User::ProviderDirectoryservice
```

But those values come up `nil` if I'm using an OS.

```
The provider is
the class is NilClass
```

Looking through the code shouldn't it be grabbing a, uh, `useradd` provider in place of the `directoryservice` provider when a different operating system is specified?

rodjek:
This should now be resolved by #498

howdoicomputer:
oh wow what a blast from the past thanks @rodjek

Issue #376, opened 2016-03-21
Provider code called in rspec testing


cdenneen:
[PUP-6067](https://tickets.puppetlabs.com/browse/PUP-6067) exposed 2 things:
1. The pip3 upgrade actually wasn't complete and caused issues in spec testing on OS X for RHEL specific code.
2. The pip, first provider, fallback was actually masking the fact that rspec-puppet is using that to mock out some facts about providers. When patch for the above bug gets merged into 4.4.1 it will be hidden again but issue is should rspec-puppet be calling that provider code. 
When compiling a catalog on master (Linux) it shouldn't be calling and looking for binaries appropriate for another platform, therefore testing on Mac shouldn't be looking for /bin/rpm, /bin/yum, /bin/mount, /bin/umount.

I think this is a long time bug exposed with the pip3 upgrade incompleteness.

DavidS:
Hi, are you sure you are in the right repo? I do not understand what you are talking about, and PUP-6167 does not exist.

cdenneen:
@DavidS yes, updated the link

DavidS:
This seems to be a variation of #352 . Besides deeply intrusive mocks that disconnect most of puppet from the host system, there is not much that can be done at this point in time.

cdenneen:
@DavidS can you explain how is this a variation of #352, that seems to be literally a puppet_gem install and test. 

This is a package resource like:

```puppet
package {'jenkins':
  ensure => '1.6.23-1',
}
```
since pip was the first provider on my system before and supported versionable this worked.
Now with 4.2.0 -> 4.4.0 pip provider isn't supporting versionable (this is being fixed in that bug ticket).
So without it rspec-puppet is looking for /bin/rpm or /bin/yum on the system to use that package provider in order to determine if it's versionable supporting or not.
I believe the facts supplied for kernel, osfamily, operatingsystem, etc in the test should be enough to determine which package provider to use vs looking for the binaries of them. As you said stubbing out these for defaultprovider isn't a long term solution because then you are left stubbing out things like /bin/mount and /bin/umount for mount {'..':} resources, and so on.

DavidS:
> I believe the facts supplied for kernel, osfamily, operatingsystem, etc in the test should be enough to determine which package provider to use vs looking for the binaries of them.

Emphasis on "should". Puppet is not implemented this way. rspec-puppet currently has no mechnisms in place working around this globally, and it would be a massive effort to do so.

[...]

Issue #256, opened 2015-02-11
compile matcher doesn't honor facts when selecting a default provider


geoffdavis:
Possibly related to GH-175, the compile matcher doesn't seem to take into account overridden facts when deciding which provider to use. I discovered this when testing a class that made use of the User type.

On my Mac, no matter what the osfamily fact was set to, the provider for Users was set to the OSX default of "directoryservice". Other matchers seem to work fine, such as "contain_user".

My workaround has been to set a pre_condition that overrides the automatic provider selection on a per-OS basis.

Here's a simple example that tests provider specific behavior. Running this on my Mac results in the Mac's directoryservice provider being used for the User type, despite the facts being set on the context.

``` puppet
class users::test {
  # Darwin cannot use managehome by default with directoryservice provider
  $my_managehome = $::osfamily ? {
    'Darwin' => false,
    default => true,
  }
  file { '/home': ensure => directory }
  user { 'testuser':
    managehome => $my_managehome,
    home => '/home/testuser',
  }
}
```

``` ruby
require 'puppetlabs_spec_helper/module_spec_helper'
describe 'users::test', :type => :class do
  [
    { :operatingsystem => 'Darwin',
      :osfamily => 'Darwin',
      :kernel => 'Darwin',
    },
    { :operatingsystem => 'CentOS',
      :osfamily => 'RedHat',
      :kernel => 'Linux',
    },
    { :operatingsystem => 'Solaris',
      :osfamily => 'Solaris',
      :kernel => 'SunOS',
    },
  ].each do |f|
    context "On #{f[:operatingsystem]}" do
      let(:facts) { f }
      it { should compile } # <-- this fails on the Solaris and Linux cases when run on a Mac
      it { should contain_user('testuser') } # <-- this works all all three platforms
    end
  end  
end
```

When these tests are run on a Mac, I get output like this:

```
$ rspec --color --format d spec/classes/users_test_spec.rb

users::test
  On Darwin
    should compile into a catalogue without dependency cycles
    should contain User[testuser]
  On CentOS
    should compile into a catalogue without dependency cycles (FAILED - 1)
    should contain User[testuser]
  On Solaris
    should compile into a catalogue without dependency cycles (FAILED - 2)
    should contain User[testuser]

Failures:

  1) users::test On CentOS should compile into a catalogue without dependency cycles
     Failure/Error: it { should compile } # <-- this fails on the Solaris and Linux cases when run on a Mac
[...]

Issue #175, opened 2014-02-18
'should compile' doesn't like 'groups' in 'user' definitions


coofercat:
I don't think 'should compile' supports 'groups' when defining users. Here's my init.pp:

```
class example () {
  user { 'example' :
    ensure  => present,
    uid     => 1234,
    gid     => 1234,
    groups  => 'localaccounts',
  }
}
```

Here's my init_spec.rb:

```
require 'spec_helper'

describe 'example' do
  context 'see if it compiles' do
    it { should compile }
  end
end
```

Here's the output of rspec -b:

```
F

Failures:

  1) example see if it compiles should compile into a catalogue without dependency cycles
     Failure/Error: it { should compile }
     NoMethodError:
       undefined method `groups' for nil:NilClass
     # /usr/local/lib/ruby/gems/2.0.0/gems/puppet-3.4.2/lib/puppet/property/list.rb:57:in `retrieve'
     # /usr/local/lib/ruby/gems/2.0.0/gems/puppet-3.4.2/lib/puppet/property/list.rb:46:in `should'
     # /usr/local/lib/ruby/gems/2.0.0/gems/puppet-3.4.2/lib/puppet/type/user.rb:359:in `block (2 levels) in <module:Puppet>'
     # /usr/local/lib/ruby/gems/2.0.0/gems/puppet-3.4.2/lib/puppet/type.rb:1995:in `instance_eval'
     # /usr/local/lib/ruby/gems/2.0.0/gems/puppet-3.4.2/lib/puppet/type.rb:1995:in `block in autorequire'
     # /usr/local/lib/ruby/gems/2.0.0/gems/puppet-3.4.2/lib/puppet/type.rb:1973:in `block in eachautorequire'
     # /usr/local/lib/ruby/gems/2.0.0/gems/puppet-3.4.2/lib/puppet/type.rb:1972:in `each'
     # /usr/local/lib/ruby/gems/2.0.0/gems/puppet-3.4.2/lib/puppet/type.rb:1972:in `eachautorequire'
     # /usr/local/lib/ruby/gems/2.0.0/gems/puppet-3.4.2/lib/puppet/type.rb:1990:in `autorequire'
     # /usr/local/lib/ruby/gems/2.0.0/gems/puppet-3.4.2/lib/puppet/graph/relationship_graph.rb:168:in `block in build_autorequire_dependencies'
[...]

PR #393, opened 2016-07-05
add mocks to allow cross os testing

tuxmea:
taken from
https://github.com/maestrodev/maestro-puppet-example/blob/master/spec/spec_helper.rb

work around https://tickets.puppetlabs.com/browse/PUP-1547
Ticket mentions fixed in 3.6.0 but rspec testing e.g. on OSX still
throws errors for package ensure set to latest and user with managehome
attribute and exec with user set to not root

should this be configurable?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,12 @@ end
 
 gemspec
 
-gem 'rake'
+if RUBY_VERSION < '1.9.3'
+  gem 'rake', '< 11'
+  gem 'json_pure', '< 2.0'
+else
+  gem 'rake', :require => false
+end
 gem 'rspec', *location_for(ENV['RSPEC_GEM_VERSION'] || '~> 3.0')
 gem 'puppet', *location_for(ENV['PUPPET_GEM_VERSION'] || '~> 4.0')
 
--- a/Gemfile
+++ b/Gemfile
@@ -12,12 +12,7 @@ end
 
 gemspec
 
-if RUBY_VERSION < '1.9.3'
-  gem 'rake', '< 11'
-  gem 'json_pure', '< 2.0'
-else
-  gem 'rake', :require => false
-end
+gem 'rake'
 gem 'rspec', *location_for(ENV['RSPEC_GEM_VERSION'] || '~> 3.0')
 gem 'puppet', *location_for(ENV['PUPPET_GEM_VERSION'] || '~> 4.0')
 
--- a/lib/rspec-puppet/spec_helper.rb
+++ b/lib/rspec-puppet/spec_helper.rb
@@ -14,7 +14,7 @@ RSpec.configure do |c|
       require 'puppet/provider/confine/exists'
       Puppet::Provider::Confine::Exists.any_instance.stubs(:which => '')
     else
-      # ONLY WORKING WITH PUPPET > 3.2 !!
+      # ONLY WORKING WITH PUPPET >= 3.2 !!
       require 'puppet/confine/exists'
       Puppet::Confine::Exists.any_instance.stubs(:which => '')
     end
--- a/lib/rspec-puppet/spec_helper.rb
+++ b/lib/rspec-puppet/spec_helper.rb
@@ -6,4 +6,19 @@ RSpec.configure do |c|
   c.module_path = File.join(fixture_path, 'modules')
   c.manifest_dir = File.join(fixture_path, 'manifests')
   c.environmentpath = File.join(Dir.pwd, 'spec')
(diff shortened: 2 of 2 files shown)


tuxmea:
Seems as if this also is a fix for
#387 #376 #256 #175

daenney:
This looks sane to me. Can you add an example or two to further elaborate on what it solves? I know you linked some issues but for completeness it would be nice to see the use cases here.

Gemfile
```diff
@@ -12,7 +12,12 @@ end
 
 gemspec
 
-gem 'rake'
+if RUBY_VERSION < '1.9.3'
```
bastelfreak: instead of this, could we drop ruby1.8 support? This is so old and EOL since a long time. voxpupuli started to drop ruby1.8 for each new module release a few months back.

tuxmea: My 5pc: There are still many Puppet 2.6, 2.7 on ruby 1.8 installations around the planet. Keeping at least rspec-puppet and puppet-lint with ruby 1.8 support will allow these people to easier migrate to newer versions of puppet. YMMV.
I suppose that dropping ruby 1.8 would require a major version bump.

rnelson0: Yes, dropping ruby 1.8 requires a new major version, and at least one minor version that announces the support will be deprecated.

As to whether we should... I'm not certain that using 1.8.7 will allow all the modern tooling to work anyway. I'd announce deprecation now and move to drop it shortly. Puppet 3 will be EOS in December, after all.

rnelson0: I also think that this deprecation should be in a separate commit rather than tied to mock changes.

lib/rspec-puppet/spec_helper.rb
```diff
@@ -6,4 +6,19 @@
...
   c.environmentpath = File.join(Dir.pwd, 'spec')
+  c.before(:each) do
+    # work around https://tickets.puppetlabs.com/browse/PUP-1547
+    # ensure that there's at least one provider available by emulating that any
+    if Puppet::version < '3.2'
+      # ONLY WORKING WITH PUPPET < 3.2 !!
+      require 'puppet/provider/confine/exists'
+      Puppet::Provider::Confine::Exists.any_instance.stubs(:which => '')
+    else
+      # ONLY WORKING WITH PUPPET > 3.2 !!
```
rnelson0: `>= 3.2` right?

rnelson0:
Ah, I see why you wanted to kill 1.8.7. I think in a separate PR, you should do that and mark 1.8.7 as allowable failures, then rebase this on top of it.

tuxmea:
More descriptions:
when not (!) using RspecPuppetFacts

1. Package declaration:
```
  package { $::demo::package_name:
    ensure => latest,
  }
```
will result on OS X to:
```
  1) demo supported operating systems on Debian demo class without any parameters should compile into a catalogue without dependency cycles
     Failure/Error: it { is_expected.to compile.with_all_deps }
       error during compilation: Parameter ensure failed on Package[demo]: Provider must have features 'upgradeable' to set 'ensure' to 'latest' at /Users/mea/demo/spec/fixtures/modules/demo/manifests/install.pp:9
     # ./spec/classes/example_spec.rb:16
```
Test Code only sets osfamily fact:
```
      os = 'Debian'
      context "on #{os}" do
        let(:facts) do
          {
            :osfamily => 'Debian',
          }
        end

        context "demo class without any parameters" do
          it { is_expected.to compile.with_all_deps }
```

2. User declaration
```
  user { 'demo':
    ensure     => present,
    managehome => true,
    groups     => ['foo', 'bar'],
  }
```
will result on OS X to:
```
  1) demo supported operating systems on Debian demo class without any parameters should compile into a catalogue without dependency cycles
     Failure/Error: it { is_expected.to compile.with_all_deps }
       error during compilation: Parameter managehome failed on User[demo]: User provider directoryservice can not manage home directories at /Users/mea/demo/spec/fixtures/modules/demo/manifests/install.pp:16
     # ./spec/classes/example_spec.rb:16
```

tuxmea:
ruby 1.8.7 is in #394

binford2k:
If this eventually leads to me not needing crap like this:

```ruby
# fake out the file checks so that they validate as absolute even though they're Windows paths
before :each do
  Puppet[:autosign] = false
  Puppet::Util::Platform.stubs(:windows?).returns true
end
```

then I'll buy you many beers at puppetconf.

tuxmea:
Code is in the wrong file. I am still trying to get this working directly in rspec-puppet.

Issue #322, opened 2015-09-28
site.pp file not being sourced by puppet during catalog compilation


dudemcbacon:
Hello,

I'm seeing the following error messages:

```
profile::ipa::client
  with default for all parameters
    should contain Notify[this is from site.pp] (FAILED - 1)

Failures:

  1) profile::ipa::client with default for all parameters should contain Notify[this is from site.pp]
     Failure/Error: it { should contain_notify('this is from site.pp') }
       expected that the catalogue would contain Notify[this is from site.pp]
```

I have the following in my site.pp file (at spec/fixtures/manifests/site.pp):

```
notify { 'this is from site.pp': }
```

I am using the puppetlabs_spec_helper to help configure RSpec. This setup worked fine before our company transitioned to Puppet 4 but now it seems like anything put into the site.pp file is ignored.

This is the content of my spec_helper.rb (at spec/spec_helper.rb) file:
```
require 'puppetlabs_spec_helper/module_spec_helper'
```

This is the content of the specfile (at spec/classes/profile__ipa__client_spec.rb):
```
require 'spec_helper'

describe 'profile::ipa::client' do
  context 'with default for all parameters' do
    it { should contain_notify('this is from site.pp') }
  end
end
```

dudemcbacon:
I created a test puppet module to better isolate this failure from the other code in the module here: https://github.com/dudemcbacon/testmodule

bundle install --path local/vendor
bundle exec rake spec

Running the above commands from that repository yields:

```
brandon ➜  testmodule git:(master) bundle exec rake spec
/home/brandon.burnett/.rvm/rubies/ruby-2.1.5/bin/ruby -I/home/brandon.burnett/development/testmodule/local/vendor/ruby/2.1.0/gems/rspec-core-3.2.3/lib:/home/brandon.burnett/development/testmodule/local/vendor/ruby/2.1.0/gems/rspec-support-3.2.2/lib /home/brandon.burnett/development/testmodule/local/vendor/ruby/2.1.0/gems/rspec-core-3.2.3/exe/rspec --pattern spec/\{classes,defines,unit,functions,hosts,integration\}/\*\*/\*_spec.rb --color
.F

Failures:

  1) testmodule with default for all parameters should contain Notify[this is from site.pp]
     Failure/Error: it { should contain_notify('this is from site.pp') }
       expected that the catalogue would contain Notify[this is from site.pp]
[...]

PR #394, opened 2016-07-06
allow working ruby 1.8 builds - make failed ruby 1.8 builds not break complete build

tuxmea:
make ruby 1.8 builds on travis to allow failures

PR #399, opened 2016-08-04
Fix: site.pp content is not being imported

jk2l:
Patch for the issue mentioned in https://github.com/rodjek/rspec-puppet/issues/322#issuecomment-237459594

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,9 +12,14 @@ end
 
 gemspec
 
-gem 'rake'
+if RUBY_VERSION <= '1.8.7'
+  gem 'rake', '<= 10.5.0'
+else
+  gem 'rake'
+end
 gem 'rspec', *location_for(ENV['RSPEC_GEM_VERSION'] || '~> 3.0')
 gem 'puppet', *location_for(ENV['PUPPET_GEM_VERSION'] || '~> 4.0')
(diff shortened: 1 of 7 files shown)


rnelson0:
@jk2l You'll need to add https://github.com/voxpupuli/puppet-community_kickstarts/blob/master/Gemfile#L30 to the Gemfile to fix the remaining errors, I believe.

jk2l:
@rnelson0 patching the Gemfile is not a problem, but i am not sure do the repo owner have their own release guideline for this repo...

rnelson0:
I think it would be acceptable to add that to this PR, as long as it's in a separate commit.

jk2l:
@rnelson0 It seem it still break for 1.8.7 + Puppet 3. Not sure how that part fix

rnelson0:
I have no idea, that makes no sense to me. Perhaps @DavidS may have a suggestion?

jk2l:
oh that seem to be fixed by https://github.com/rodjek/rspec-puppet/pull/394 so just need to wait for that PR pull into master and then do a rebase and merge

DavidS:
I just sorted out the Gemfile, so it works again on all supported platforms. If you could rebase, we could see how the code is faring now.

Thanks for your patience.

DavidS:
Also, wow me, for miscklicking >.<

Apologies for the noise.

Issue #400, opened 2016-08-05
Incorrect trusted node data


jk2l:
Trusted node data should have $trusted['certname'] that is like fqdn. curretnly it is being hardcoded by Puppet::Test::TestHelper as "testing"

I think this one should be set properly

jk2l:
Note: the current trusted data in rspec break test case that depends on trusted data

rodjek:
Looks like #444 has been merged in which also implements this, so I'll close this PR. Thank you for the contribution though!

rodjek:
This should be resolved by #444, please reopen if it's still not working for you

PR #401, opened 2016-08-05
Fix: incorrect trusted node data

jk2l:
https://github.com/rodjek/rspec-puppet/issues/400

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -24,6 +24,13 @@ module RSpec::Puppet
       end
 
       node_name = nodename(type)
+      Puppet.push_context(
+        {
+          :trusted_information =>
+            Puppet::Context::TrustedInformation.new('local', node_name, {}),
+        },
+        "Context for specs"
+      )
 
       hiera_config_value = self.respond_to?(:hiera_config) ? hiera_config : nil
 


hlindberg:
A `push_context` is in itself not bad, but it really needs to be balanced with a `pop_context`, unless naturally there is just setup for a single run that then exits the process.
If it keeps piling on contexts (i.e. never pops), then in it is bad as it will consume memory and gradually make context lookups slower.
I could not determine if the `push_context` in this PR is bad or not - would need to understand a lot more than what I could quickly glean from the source around where it appears.

jk2l:
The code basically copy from https://github.com/puppetlabs/puppet/blob/ed209118fceb6433b51c88c10883cb7415b36adb/lib/puppet/test/test_helper.rb i am not entirely sure how the logic flow internally

DavidS:
I've re-read the code, and this should wrap the `super` call at https://github.com/rodjek/rspec-puppet/blob/master/lib/rspec-puppet/adapters.rb#L148 . That will make it puppet4 only, and allow you to unconditionally use something like 

```
result = nil
Puppet.override({ :trusted_information ... }, "rspec-puppet test scope") do
  result = super
end
result
```

as that code only runs on puppet 4 by design.

jk2l:
So what is the approach of fixing trusted node data support for rspec?

DavidS:
@jk2l it'd be great if you could test out if the approach I described in my earlier comment works for your test case.

rodjek:
Looks like #444 has been merged in which also implements this, so I'll close this PR. Thank you for the contribution though!

PR #408, opened 2016-08-26
Make :undef spec more expressive. Closes #407

### LINKED ISSUE #407: Passing param :undef is broken
Using the below fixture and spec tests, the last test will fail.

**Problem:** Instead of passing the value `:undef`, the attribute `group` just gets assigned its default value `'users'`.

```Puppet
# spec/fixtures/modules/undef_test/manifests/init.pp
class undef_test(
  $user  = undef,
  $group = 'users'
) { }
```

```Ruby
# spec/classes/undef_spec.rb
require 'spec_helper'

describe 'undef_test' do
  it { should compile.with_all_deps }

  context 'with params unset' do
    it { should contain_class('undef_test').without_user }

    it { should contain_class('undef_test').with(group: 'users') }
  end

  context 'with group => undef' do
    let(:params) { { :group => :undef } }

    it { should contain_class('undef_test').without_user }

    it { should contain_class('undef_test').without_group }
  end
end
```

leoarnold:


leoarnold:
Failed Travis CI checks are not due to code, but a gem dependency problem on Ruby 1.9

DavidS:
Yeah, I'm looking into the latter currently. I'll rekick travis here once
I've found all necessary magic incantations.

DavidS:
Close/re-open to rekick travis with current master.

DavidS:
@leoarnold puppet 2.7 has some issues with your code :-/

leoarnold:
@DavidS As noted in #407, the only meaningful use of passing `:undef` is to block out required attributes. In any other case, passing `:undef` is just the same as not mentioning that attribute in the `:params` hash.

Obviously, Puppet 2.7 does not support blocking out required attributes using `:undef`, rendering this functionality useless in Puppet 2.7.

Not sure how to deal with this.

DavidS:
You can make the specs conditional on the puppet version, see https://github.com/rodjek/rspec-puppet/blob/master/spec/classes/array_spec.rb#L43 for an example.

leoarnold:
@DavidS Amended and extended. Thanks for the hint.

PR #411, opened 2016-09-01
(ORCH-1518) Introduce app_management testing support

kbarber:
This patch adds support the `app_management` feature added to Puppet in 4.3.

    It provides a new type `application` that can now be created much like a class
    or resource. It also enables the `app_management` setting automatically when
    the version of Puppet provided has this feature.

    To assist with passing resource references as required by application node mappings,
    this patch introduces the new `ref` helper to allow test writers to provide
    resource refs that appear as raw strings (instead of being quoted like normal strings).

    All this new support is documented in the README.md, and tests have been added
    to provide basic coverage for this new work.

    Signed-off-by: Ken Barber <ken@bob.sh>

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ structure and naming convention.
            |     |
            |     +-- <define_name>_spec.rb
            |
+           +-- applications
+           |     |
+           |     +-- <application_name>_spec.rb
+           |
            +-- functions
            |     |
            |     +-- <function_name>_spec.rb
@@ -64,6 +68,10 @@ describe 'mydefine', :type => :define do
   ...
 end
 
+describe 'myapplication', :type => :application do
+  ...
+end
+
 describe 'myfunction', :type => :puppet_function do
   ...
 end
@@ -77,7 +85,7 @@ describe 'myhost.example.com', :type => :host do
 end
 ```
 
-## Defined Types & Classes
+## Defined Types, Classes & Applications
 
 ### Matchers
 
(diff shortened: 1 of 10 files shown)


kbarber:
Todo:

- Needs puppet4 only guard
- Automatic app_management config setting

kbarber:
For a working implementation, see here: https://github.com/puppetlabs/puppetlabs-wordpress_app/pull/13

kbarber:
@nicklewis @DavidS @adreyer WIP patch for basic testing.

README.md
```diff
@@ -450,6 +477,49 @@ You can also use `exported_resources` directly in a test:
...
+require a number of elements already documented above to be combined for testing.
+
+A full example of the simplest rspec test for a single component application:
+
+```ruby
+require 'spec_helper'
+
+describe 'orch_app' do
+  before(:each) do
+    Puppet[:app_management] = true
```
DavidS: Wouldn't you want this enabled at all times? Is there a reason to not have it enabled?

spec/fixtures/modules/orch_app/manifests/db.pp
```diff
@@ -0,0 +1,3 @@
+define orch_app::db(
+) {
+}
```
DavidS: ^^

README.md
```diff
@@ -450,6 +477,49 @@ You can also use `exported_resources` directly in a test:
...
+  end
+
+  it { should compile }
+end
+```
+
+Each piece is required:
+
+* You must turn on app_management during testing for the handling to work
+* The `:node` definition is required to be set so later on you can set in it `:params`
```
DavidS: s/can set/can reference ;

lib/rspec-puppet/example/application_example_group.rb
```diff
@@ -0,0 +1,19 @@
+module RSpec::Puppet
+  # This module provides support for the application type
+  module ApplicationExampleGroup
```
DavidS: It hurts a little bit that this is an exact copy of the ClassExampleGroup

spec/support_spec.rb
```diff
@@ -26,4 +26,10 @@
       expect($LOAD_PATH).to include(dirb)
     end
   end
+
+  describe "#ref" do
+    it 'should return a new RawString with the type/title format' do
+      expect(subject.ref('Package','tomcat').inspect).to eq("Package['tomcat']")
```
DavidS: You're testing too much here.

```
it { expect(subject.ref("A", "b").to be_a RSpec::Puppet::RawString }
```

is enough. RawString's behaviour is tested there.

DavidS: ah, formatting is happening here too. Makes your test more appealing, yes.

DavidS:
Regarding docs: if you could add a usage examples for `ref()` outside the app orch section, that would be great to show that off.

README.md
kbarber: Its on my todo list already.

lib/rspec-puppet/example/application_example_group.rb
kbarber: And HostExampledGroup and DefineExampleGroup.

spec/fixtures/modules/orch_app/manifests/db.pp
kbarber: fixed.

README.md
kbarber: Fixed.

spec/support_spec.rb
kbarber: Cool, so I'll leave it then.

README.md
kbarber: fixed.

lib/rspec-puppet/example/application_example_group.rb
kbarber: I'm going to leave this for now, I imagine it should be easy to solve whenever, just don't want to add such a refactor right now to this PR if thats okay.

kbarber:
@DavidS added more examples to docs also. If you wanted to take another pass at reviewing this, it would be appreciated. I've taken off the WIP marker.

lib/rspec-puppet/example/application_example_group.rb
DavidS: :+1: \*sigh*

spec/support_spec.rb
DavidS: yup

PR #412, opened 2016-09-02
(MODULES-3737) Allow testing internal puppet 4 functions

DavidS:
Some puppet 4 functions require access to the scope, which needs to be
provided through the loader.

Since the loader override needs to be active during function evaluation,
this also adds a wrapper for that func, and changes so that the func name
is displayed properly.

To reduce complexity of the rest of the code, a similar wrapper is
introduced for puppet 3. This makes the rest of the code simpler.

### DIFF
--- a/lib/rspec-puppet/example/function_example_group.rb
+++ b/lib/rspec-puppet/example/function_example_group.rb
@@ -7,23 +7,23 @@ module RSpec::Puppet
     def subject
       function_name = self.class.top_level_description.downcase
 
-      vardir = setup_puppet
+      with_vardir do
+        if Puppet.version.to_f >= 4.0
+          env = adapter.current_environment
+          loader = Puppet::Pops::Loaders.new(env)
+          func = loader.private_environment_loader.load(:function, function_name)
+          return func if func
+        end
 
-      if Puppet.version.to_f >= 4.0
-        env = adapter.current_environment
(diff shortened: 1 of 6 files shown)


spec/unit/matchers/run_spec.rb
```diff
@@ -1,19 +1,14 @@
 require 'spec_helper'
 
 describe RSpec::Puppet::FunctionMatchers::Run do
-  it 'should call the lambda with no params' do
+  fit 'should call the lambda with no params' do
```
hunner: errrrt

lib/rspec-puppet/example/function_example_group.rb
```diff
@@ -4,27 +4,63 @@ module FunctionExampleGroup
...
+      attr_accessor :func_name
+
+      def initialize(name, func)
+        @func_name = name
+        @func = func
+      end
+
+      def call(*args)
+        if args.nil?
+          @func.call
```
hunner: Would this not be `@func.call([])`?

hunner: Oh never mind, I see it below.

DavidS:
See https://github.com/puppetlabs/puppetlabs-stdlib/pull/649 for example usage.

PR #422, opened 2016-10-27
Fix naming of test_api specs so they run

domcleal:
Incorrectly named ever since they were added 7dddebd, updated to take
into account the change of `subject` in 34220fc and rspec should/expect
style changes.


DavidS:
mergebot at your service :-)

thanks a lot for your help!

PR #423, opened 2016-10-27
Add type_alias example group and allow_value(s) matchers

### LINKED ISSUE #402: Support testing variable types
Now that you can put types into `module/types/mytype.pp`, it would be nice if rspec-puppet could support testing them by passing in various values and checking the type matches / does not match.

domcleal:
New allow_value and allow_values matchers can test type aliases defined
on Puppet 4.4+ to check that they do/don't accept certain values by
building a test catalog with Puppet's assert_type function.

These are available for specs with type :type_alias, under the directory
specs/type_aliases/ that describe the alias name (e.g. My::TypeAlias).

Fixes #402

---

This should help replace [use of test classes](https://github.com/puppetlabs/puppetlabs-stdlib/commit/83539371e5b1e6ff412eda48c3a725653c506ece#diff-f35de9572e500eb5a1f229d42f9d69a4R4) with shorter specs that need much less boilerplate.


DavidS:
Ugh, I must've missed that last year :-( Apologies!

Is there any restriction to type *aliases*?  The top-level term used in puppet is "data type" (see https://docs.puppet.com/puppet/latest/lang_data_type.html).

domcleal:
> Is there any restriction to type aliases? The top-level term used in puppet is "data type" (see https://docs.puppet.com/puppet/latest/lang_data_type.html).

No restriction really, I've been a bit unsure what to call them. I think data types cover the built in types as well - the ability to put data types under module/types/ is specifically a type alias: [puppet-specifications: Type Alias Expression](https://github.com/puppetlabs/puppet-specifications/blob/master/language/expressions.md#type-alias-expression).

stdlib used spec/aliases/ too, and while I'd like to mirror modules/types/ under spec/types/, that's already used for resource type examples. I stuck with type_aliases as I thought it bridged the module dir name and the specification name.

Issue #424, opened 2016-11-21
RSpec::Puppet::Coverage.add_filter has incorrect logic


shadow999:
I tried to exclude my `class foo::bar` by calling `RSpec::Puppet::Coverage.add_filter('class', 'foo::bar')`. But it actually added `Class[Foo::bar]` into the array. This will not work, since catalog contains `Class[Foo::Bar]`. 

I am wondering if `add_filter` can detect if it is class type then capitalize the first letter of the entire string and also the first letter after `::`?

shadow999:
#425

DavidS:
Merged #425

PR #425, opened 2016-11-22
Fixed RSpec::Puppet::Coverage.add_filter capitalize logic

shadow999:
- First character of whole name and first character after :: should be capitalized. This rule should be applied to both type name and resource title.


shadow999:
#424

PR #443, opened 2017-01-25
Bugfix/puppet4 freeze parameters

### LINKED ISSUE #442: puppet 4 freezes arguments for a function
Puppet error: `Error while evaluating a Function Call, can't modify frozen hash `

Rspec-puppet should also freeze all parameters before passing them through to the function. Only after adding manual .freeze on the parameters, I'm able to reproduce the same error while running rspec tests

vStone:
Fixes #442


DavidS:
Ha! Good find. Thanks a lot.

PR #444, opened 2017-01-25
Add full support for the trusted facts hash

seanmil:
Set the standard trusted fact keys based on the 'node' name

Add 'trusted_facts' option to allow setting custom
trusted fact extensions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,6 +519,31 @@ However, if you want to specify it in each example, you can do so
 let(:module_path) { '/path/to/your/module/dir' }
 ```
 
+#### Specifying trusted facts
+
+When testing with Puppet >= 4.3 the trusted facts hash will have the standard trusted fact keys
+(certname, domain, and hostname) populated based on the node name (as set with `:node`).
+
+By default, the test environment contains no custom trusted facts (as usually obtained
+from certificate extensions) and found in the `extensions` key. If you need to test against
+specific custom certificate extensions you can set those with a hash. The hash will then
+be available in `$trusted['extensions']`
+
+```ruby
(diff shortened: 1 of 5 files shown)


seanmil:
This is intended to address #400

README.md
```diff
@@ -519,6 +519,29 @@ However, if you want to specify it in each example, you can do so
...
+#### Specifying trusted certificate extensions
+
+In Puppet 4 the trusted facts hash will have the standard keys (certname, domain, and hostname) populated
+based on the node name (as set with `:node`).
+
+By default, the test environment contains no custom trusted extensions. If you need to test against specific
+certificate extensions you can set those with a hash
+
+```ruby
+let(:trusted_fact_extensions) { {'pp_uuid' => 'ED803750-E3C7-44F5-BB08-41A04433FE2E', '1.3.6.1.4.1.34380.1.2.1' => 'ssl-termination'} }
```
DavidS: Would you be open to changing this to just `trusted_facts`? I see no reason to concern users with the implementation details.

spec/classes/trusted_facts_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'spec_helper'
+
+describe 'trusted_facts', :if => Puppet.version.to_f >= 4.0 do
```
DavidS: Given the test results, and the fact that 4.0, and 4.1 are not supported anymore, you can push this check up to `>= 4.3` and mention this in the readme.

README.md
seanmil: I debated quite a while on what to call that. No issues with it. The update I just pushed includes the change to `trusted_facts`.

spec/classes/trusted_facts_spec.rb
seanmil: Done and done. Thanks!

DavidS: I'll merge it as soon as travis is OK

PR #445, opened 2017-01-26
Extend trusted fact support to function testing

seanmil:
Add support for testing with trusted facts within functions.


seanmil:
@DavidS Sorry I didn't have this included on the first go-around for #400 . It didn't originally occur to me that custom functions would be tested in a different context than classes. Based on what I see in lib/rspec-puppet/example I believe that functions were the only gap. Thanks!

DavidS:
Don't feel guilty about it. I could have remembered too.

It's fixed now, so no harm done.

PR #446, opened 2017-01-27
Do not freeze classes passed into functions

domcleal:
Freezing a class prevents it from being modified at all, which can
happen in unit tests with mocha or rspec-expections.

Calling a function with `String` as an argument will freeze the entire
class and prevent `allow_any_instance_of(String)...` type expectations
from adding their hooks to the class.

e.g. stdlib:
  https://github.com/puppetlabs/puppetlabs-stdlib/blob/00c881d/spec/functions/is_a_spec.rb#L18
  https://github.com/puppetlabs/puppetlabs-stdlib/blob/00c881d/spec/functions/pw_hash_spec.rb#L57

### DIFF
--- a/spec/fixtures/modules/test/lib/puppet/functions/frozen_function.rb
+++ b/spec/fixtures/modules/test/lib/puppet/functions/frozen_function.rb
@@ -1,5 +1,5 @@
 Puppet::Functions.create_function(:frozen_function) do
   def frozen_function(value)
-    value.reverse!
+    value.frozen?
   end
 end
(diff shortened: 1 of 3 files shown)


spec/fixtures/modules/test/lib/puppet/functions/frozen_function.rb
```diff
@@ -1,5 +1,5 @@
 Puppet::Functions.create_function(:frozen_function) do
   def frozen_function(value)
-    value.reverse!
+    value.frozen?
```
vStone: This kinda changes what is being tested: You will most likely in a real life scenario not check for frozen but rather by accident execute a method on an object that changes it.

domcleal: Sure, but this is a unit test to check the arguments are frozen rather than a test of `reverse!` or other methods that actually modify an object. Checking if the object's `#frozen?` is equivalent to calling a method that raises an exception `if frozen?`.

Issue #440, opened 2017-01-23
Unable to test conditional that uses $::environment with puppet 3.8 and 4.3


italy-portugal:
Example:

``` puppet
class test {
  notify { $::environment: }
  if $::environment == 'testenv' {
      notify { 'I can test environments': }
  }
}
```

``` ruby
require 'spec_helper'
describe 'test' do
  let(:environment) { 'testenv' }
  
  # This one succeeds
  it { is_expected.to contain_notify('testenv') }

  # This one fails
  it { is_expected.to contain_notify('I can test environments') }
end
```

trzejos:
I have had similar issues with $;;environment as well. Not exactly a fix for this bug, but you can try replacing $::environment == 'testenv' with $::environment =~ /^testenv$/ in your manifest.

DavidS:
#453 has a minimal reproduction, and it does indeed fail on 3.8, and 4.3, but works on a range of other puppet versions.

domcleal:
Just to be clear, the merged PR will fix it on 4.0-4.3, but it will continue to fail on 3.8.5 to .7. Apart from these few patch versions (which unfortunately are the latest 3.x), it should now work reliably.

italy-portugal:
Thanks for the clarification!

PR #453, opened 2017-02-02
(WIP) Add environment testing

DavidS:
Reproduces #440


DavidS:
re-triggering travis with correct metadata

DavidS:
I just noticed that my example fails on more versions than your example, which I initially tested. I'll look into the details tomorrow.

domcleal:
#456 contains a fix for Puppet 4.0 to 4.3, but I wasn't able to fix it under 3.8.5 to .7.

PR #456, opened 2017-02-07
Fix environment override on 4.x (#440)

### LINKED ISSUE #440: Unable to test conditional that uses $::environment with puppet 3.8 and 4.3

domcleal:
Replaces #453 to fix #440 on 4.0 to 4.3, unable to fix it neatly on 3.8.5 to .7 though.

---

**Override $::environment parameter on Puppet 4.0 - 4.3**

Prevents environment parameter being set to a symbol, which cannot
otherwise be compared correctly to strings in manifests. Fixed by
puppetlabs/puppet@0296d2d.

Fixes #440 

---

**Disable environment override test on Puppet ~> 3.8.5**

Broken since PUP-5522 (puppetlabs/puppet@8136aff), without a clear
solution.

### DIFF
--- a/lib/rspec-puppet/adapters.rb
+++ b/lib/rspec-puppet/adapters.rb
@@ -146,6 +146,9 @@ module RSpec::Puppet
 
       def catalog(node, exported)
         node.environment = current_environment
+        # Override $::environment to workaround PUP-5835, where Puppet otherwise
+        # stores a symbol for the parameter
+        node.parameters['environment'] = current_environment.name.to_s if node.parameters['environment'] != node.parameters['environment'].to_s
         super
       end
 
--- a/spec/hosts/environment_spec.rb
+++ b/spec/hosts/environment_spec.rb
@@ -4,7 +4,8 @@ describe 'facts.acme.com' do
   context 'without an explicit environment setting' do
     it { should contain_file('environment').with_path('rp_env') }
   end
-  context 'when specifying an explicit environment' do
+  # Broken on ~> 3.8.5 since PUP-5522
+  context 'when specifying an explicit environment', :unless => (Puppet.version >= '3.8.5' && Puppet.version.to_i < 4) do
     let(:environment) { 'test_env' }
     it { should contain_file('environment').with_path('test_env') }
     it { should contain_file('conditional_file') }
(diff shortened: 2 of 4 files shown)


lib/rspec-puppet/adapters.rb
```diff
@@ -146,6 +146,9 @@ def settings_map
 
       def catalog(node, exported)
         node.environment = current_environment
+        # Override $::environment to workaround PUP-5835, where Puppet otherwise
+        # stores a symbol for the parameter
+        node.parameters['environment'] = current_environment.name.to_s if Puppet.version.to_f >= 4.0 && Puppet.version.to_f < 4.4
```
DavidS: This might be more robust (version independent) by saying:

```
node.parameters['environment'] = current_environment.name.to_s if node.parameters['environment'] != node.parameters['environment'].to_s
```

spec/hosts/environment_spec.rb
```diff
@@ -4,8 +4,10 @@
   context 'without an explicit environment setting' do
     it { should contain_file('environment').with_path('rp_env') }
   end
-  context 'when specifying an explicit environment' do
+  # Broken on ~> 3.8.5 since PUP-5522
+  context 'when specifying an explicit environment', :unless => (Puppet.version >= '3.8.5' && Puppet.version.to_i < 4) do
```
DavidS: :-(

lib/rspec-puppet/adapters.rb
domcleal: Thanks, I've changed it. I was tempted to keep it version-specific so it can be removed easily later, but I think your suggestion's defensive enough.

spec/hosts/environment_spec.rb
domcleal: Sorry, this stumped me without monkey patching (which perhaps belongs in puppetlabs_spec_helper?).

domcleal:
> Travis is still unhappy on 4.x, seems like some other thing in the compile matcher broke now.

I've fixed this now. The test expected a specific line number, but the fixture changed.

PR #467, opened 2017-03-06
Several fixes to function_example_group for scoping

heliocentric:
* The current implementation uses Puppet.override to set global_scope
  when a function example is called; but this does not trickle down
  into all the functions called by the example function. This results
  in strange behaviour as self.closure_scope returns a hash if @closure_scope
  is nil (which it always will be), and Puppet.lookup(:global_scope)
  returns nil (since it's never set accept during the initial call()).

* Ensure that :loaders exists so that functions like
  lookup/hiera 5 will function properly.

* Add call_function() wrapper that mimics the same api as the scope
  function, that can be used to call another function in the same scope
  as the example function; useful for validating that the example
  function performed the correct action.

### DIFF
--- a/lib/rspec-puppet/example/function_example_group.rb
+++ b/lib/rspec-puppet/example/function_example_group.rb
@@ -79,18 +79,16 @@ module RSpec::Puppet
       @subject ||= find_function
     end
 
-    def find_function
-      function_name = self.class.top_level_description
-
+    def find_function(function_name = self.class.top_level_description)
       with_vardir do
         env = adapter.current_environment
 
         if Puppet.version.to_f >= 4.0
           context_overrides = compiler.context_overrides
           func = nil
+          loaders = Puppet.lookup(:loaders)
           Puppet.override(context_overrides, "rspec-test scope") do
-            loader = Puppet::Pops::Loaders.new(env)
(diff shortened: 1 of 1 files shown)


lib/rspec-puppet/example/function_example_group.rb
```diff
@@ -21,7 +21,7 @@ def execute(*args)
       end
 
       # compatibility alias for existing tests
-      def call(scope, *args)
```
DavidS: This is part of the public API, and cannot be changed, without accounting for existing callers.

heliocentric:
@DavidS I changed the call() interface back. However in order to use scope.call_function() and have it function equivalently to non-rspec usage, I had to push loaders and global_scope into the global Puppet context in the build_compiler() function. 

This is because when running call_function() without an explicit override the scope is dropped and it uses Puppet.lookup(:global_scope) and Puppet.lookup(:loaders) to get it's own closure_scope; this causes call_function() to break in a function called within a function, etc. So by setting it globally across the board it fixes that.

Don't know if it might break rspec in other ways as the context code seems a little strange to me.

trevor-vaughan:
@DavidS Any chance of getting this in?

DavidS:
I'm still not 100% convinced on the `call_function()` pass-through, but the rest looks fine.

PR #479, opened 2017-03-14
Use versioncmp to check Puppet version for 4.10.x compatibility

domcleal:
`Puppet.version.to_f` on Puppet 4.10.0 will evaluate to `4.1`, causing
test and behavioural changes when conditionals check that the version is
equal or greater than versions such as `4.3`.

Version comparisons that are vulnerable to this have been changed to use
Puppet's versioncmp implementation, while most others only check for
for major version boundaries which is safe.

---

This (or equivalent change) ought to be _released_ too before 4.10.x to prevent problems in module tests on 4.10's release.

### DIFF
--- a/lib/rspec-puppet/adapters.rb
+++ b/lib/rspec-puppet/adapters.rb
@@ -202,15 +202,15 @@ module RSpec::Puppet
 
     def self.get
       [
-        [4.0, Adapter4X],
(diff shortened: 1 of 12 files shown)


domcleal:
4.10.0 came out yesterday, but this is unreleased - would you be able to make a release with it in soon?

Issue #437, opened 2017-01-05
Windows module validation on puppet 4 regressions


james-powis:
windows absolute file paths have been broken since 2.3.0, While there are numerous other bug reports around this issue, I figured I would share the results of my Root Cause Analysis and some of the extreme levels of hackery I have attempted to work around the issue only to have future releases cause even more hackery needed. In short the absolute filepath issue was introduced in this commit:

https://github.com/rodjek/rspec-puppet/commit/9c295b5de726128bc08acfb1efaeaf5987174563 

I don't know WHY or how to fix it at this point, so I will leave that to those of you who have more knowledge in this code base than I do.

and Now for your entertainment:

**The Investigation (backstory and hackery)**

On versions after `2.2.0` rspec puppet has required stubbing the file path to avoid getting file path is not fully qualified errors:

example of error:
```log
     Puppet::ResourceError:
       Parameter path failed on File[C:\Windows\DevOps\mssql_std_2014.zip]: File paths must be fully qualified, not 'C:\Windows\DevOps\mssql_std_2014.zip' at /Users/jrp/git/puppet/modules/mssql/spec/fixtures/modules/mssql/manifests/init.pp:44
     # /Users/jrp/.rvm/gems/ruby-2.3.0/gems/puppet-4.8.1-universal-darwin/lib/puppet/util/errors.rb:106:in `fail'
     # /Users/jrp/.rvm/gems/ruby-2.3.0/gems/puppet-4.8.1-universal-darwin/lib/puppet/type/file.rb:56:in `block (3 levels) in <top (required)>'
     # /Users/jrp/.rvm/gems/ruby-2.3.0/gems/puppet-4.8.1-universal-darwin/lib/puppet/parameter.rb:456:in `validate'
     # /Users/jrp/.rvm/gems/ruby-2.3.0/gems/puppet-4.8.1-universal-darwin/lib/puppet/parameter.rb:491:in `value='
     # /Users/jrp/.rvm/gems/ruby-2.3.0/gems/puppet-4.8.1-universal-darwin/lib/puppet/type.rb:675:in `[]='
     # /Users/jrp/.rvm/gems/ruby-2.3.0/gems/puppet-4.8.1-universal-darwin/lib/puppet/type.rb:2454:in `set_name'
     # /Users/jrp/.rvm/gems/ruby-2.3.0/gems/puppet-4.8.1-universal-darwin/lib/puppet/type.rb:2385:in `initialize'
[...]

Issue #352, opened 2016-01-14
compile.with_all_deps fails with puppet_gem provider


ProgrammingAce:
Hey folks, I can't figure out how to get compile.with_all_deps to pass when using the puppet_gem package provider. It fails with:

```
  1) testing supported operating systems on redhat-7-x86_64 testing class without any parameters should compile into a catalogue without dependency cycles
     Failure/Error: it { is_expected.to compile.with_all_deps }
       error during compilation: Parameter provider failed on Package[r10k]: Invalid package provider 'puppet_gem' at /Users/pmcdonough/code/puppet-modules/testing/spec/fixtures/modules/testing/manifests/install.pp:9
     # ./spec/classes/example_spec.rb:14:in `block (6 levels) in <top (required)>'
```

I'm running on OSX version:10.11.2
With puppet version: 4.3.1
With rspec-puppet: 2.3.0

This is my entire test file:

```
require 'spec_helper'

describe 'testing' do
  context 'supported operating systems' do
    on_supported_os.each do |os, facts|
      context "on #{os}" do
        let(:facts) do
          facts
        end
        context "testing class without any parameters" do

          it { is_expected.to compile.with_all_deps }
          it { is_expected.to contain_package('r10k').with_provider("puppet_gem") }

        end
      end
    end
  end
end
```

And this is the only non-boilerplate code in the module:

```
class testing::install {
    package { 'r10k':
        ensure => 'installed',
        provider => 'puppet_gem',
    }
}
```

The module successfully applies to real instances.

DavidS:
Looking at the [docs](https://docs.puppetlabs.com/references/latest/type.html#package-provider-puppet_gem), the puppet_gem provider is only available if there is a local puppet installation found in /opt. rspec-puppet currently has no mechanism to stub that out.

I fear this is related to the changes in #261 where the catalog was expanded using to_ral. /cc @raphink @mcanevet

raphink:
That could indeed be a side effect of to_ral I'm afraid...

DavidAwad:
Are there any fixes to this? Are there any ways to force rspec-puppet to accept stubs for custom providers? I'm having a similar problem where I'm using puppet homebrew, but rspec fails to accept it as a valid package provider and fails the tests.

rodjek:
This should now be resolved by #498

PR #481, opened 2017-03-16
Stub out provider confines that prevent running tests for different platforms

### LINKED ISSUE #387: Odd behavior on User type

rodjek:
In order to fully resolve the graph and create all the automatic relationships we have to convert the resource catalogue produced by the compiler into a RAL catalogue like that which an agent would apply.

An unfortunate side effect of this is that it causes providers to be loaded and tested for suitability. This can cause problems for people who have manifests that specify providers for platforms other than the one they're running rspec-puppet on (e.g. testing Windows manifests on a Linux host).

This PR stubs out the feature and command confinements from the providers, effectively allowing rspec-puppet to load any provider that meets the fact value confinements (e.g. the `aix` Package provider is confined to `:operatingsystem => :aix`.

Closes #387, #376, #352, #256, #175

**NB: Not ready for merge yet, there's still an issue with stubbing the windows file provider re #437**

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -265,6 +265,20 @@ module RSpec::Puppet
     end
 
     def stub_facts!(facts)
+      if facts['operatingsystem'] && facts['operatingsystem'] == 'windows'
+        Puppet.settings[:autosign] = false
+        allow(Puppet::Util::Platform).to receive(:windows?).and_return(true)
+        begin
+          require 'puppet/util/windows'
+        rescue LoadError
+          allow_any_instance_of(Kernel).to receive(:require).with(anything).and_call_original
+          allow_any_instance_of(Kernel).to receive(:require).with('puppet/util/windows').and_return(true)
+        end
+      else
+        allow(Puppet::Util::Platform).to receive(:windows?).and_return(false)
+        RSpec::Mocks.space.proxy_for(Kernel).reset
+      end
+
       Facter.flush
       Facter.reset
       facts.each do |k, v|
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -231,7 +231,11 @@ module RSpec::Puppet
       # that providers might do. This is necessary because we convert the
       # resource catalogue to a RAL catalogue in order to evaluate all the
       # automatic graph edges, which requires evaluating the providers.
-      allow_any_instance_of(Puppet::Confine::Feature).to receive(:pass?).with(anything).and_return(true)
+      if Puppet::Util::Package.versioncmp(Puppet.version, '3.0.0') < 0
+        allow_any_instance_of(Puppet::Provider::Confine::Feature).to receive(:pass?).with(anything).and_return(true)
+      else
+        allow_any_instance_of(Puppet::Confine::Feature).to receive(:pass?).with(anything).and_return(true)
+      end
       allow_any_instance_of(Puppet::Util).to receive(:which).with(anything).and_return(true)
       allow_any_instance_of(Puppet::Provider::Command).to receive(:execute).with(any_args).and_return("")
 
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -235,9 +235,9 @@ module RSpec::Puppet
         allow_any_instance_of(Puppet::Provider::Confine::Feature).to receive(:pass?).with(anything).and_return(true)
       else
         allow_any_instance_of(Puppet::Confine::Feature).to receive(:pass?).with(anything).and_return(true)
+        allow_any_instance_of(Puppet::Provider::Command).to receive(:execute).with(any_args).and_return("")
       end
       allow_any_instance_of(Puppet::Util).to receive(:which).with(anything).and_return(true)
-      allow_any_instance_of(Puppet::Provider::Command).to receive(:execute).with(any_args).and_return("")
 
       # If we're going to rebuild the catalog, we should clear the cached instance
       # of Hiera that Puppet is using.  This opens the possibility of the catalog
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -217,7 +217,23 @@ module RSpec::Puppet
       end
     end
 
+    def reset_providers
+      Puppet::Type.eachtype do |type|
+        # Unset the default provider for each type as this is cached in a class
+        # instance method and so doesn't get re-evaluated when we change facts
+        # and compile a new catalogue.
+        type.defaultprovider = nil
+      end
+    end
+
     def build_catalog_without_cache(nodename, facts_val, trusted_facts_val, hiera_config_val, code, exported)
+      # Blindly stub out any feature and command Puppet::Confine evaluations
+      # that providers might do. This is necessary because we convert the
+      # resource catalogue to a RAL catalogue in order to evaluate all the
+      # automatic graph edges, which requires evaluating the providers.
+      allow_any_instance_of(Puppet::Confine::Feature).to receive(:pass?).with(anything).and_return(true)
+      allow_any_instance_of(Puppet::Util).to receive(:which).with(anything).and_return(true)
+      allow_any_instance_of(Puppet::Provider::Command).to receive(:execute).with(any_args).and_return("")
 
       # If we're going to rebuild the catalog, we should clear the cached instance
       # of Hiera that Puppet is using.  This opens the possibility of the catalog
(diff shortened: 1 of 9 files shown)


rodjek:
This change is incompatible with Puppet 2.x, which is the cause of the build failure. 2.x builds will be removed in #482

lib/rspec-puppet/support.rb
```diff
@@ -217,7 +217,23 @@ def with_vardir
...
+    end
+
     def build_catalog_without_cache(nodename, facts_val, trusted_facts_val, hiera_config_val, code, exported)
+      # Blindly stub out any feature and command Puppet::Confine evaluations
+      # that providers might do. This is necessary because we convert the
+      # resource catalogue to a RAL catalogue in order to evaluate all the
+      # automatic graph edges, which requires evaluating the providers.
+      allow_any_instance_of(Puppet::Confine::Feature).to receive(:pass?).with(anything).and_return(true)
+      allow_any_instance_of(Puppet::Util).to receive(:which).with(anything).and_return(true)
+      allow_any_instance_of(Puppet::Provider::Command).to receive(:execute).with(any_args).and_return("")
```
DavidS: The second and third here look like they could massively mess with people's tests.

* Direct `Puppet::Util.which` stubs, probably ok:
  * https://github.com/puppetlabs/puppetlabs-firewall/blob/cbbd9781164351d38cb8391096b8817e5fa8bf9b/spec/unit/puppet/provider/iptables_spec.rb#L32
  * https://github.com/puppetlabs/puppetlabs-mysql/blob/ef7e882ebaf98d5ccd0c16891a396db9f4700bc0/spec/unit/puppet/provider/mysql_user/mysql_spec.rb#L87
  * https://github.com/puppetlabs/puppetlabs-sqlserver/blob/master/spec/unit/puppet/sqlserver_spec_helper.rb#L6

* low-level stubs for `Puppet::Util::Execution.execute`, which will be skipped (through https://github.com/puppetlabs/puppet/blob/801542e53b6f7e7d5edcaea46590139cb42a2949/lib/puppet/provider.rb#L257)
  * https://github.com/puppetlabs/puppetlabs-chocolatey/blob/41a4307ed938a2c5edc1dac78fe8b64a4f80de4e/spec/unit/puppet/provider/chocolateyconfig/windows_spec.rb#L234
  * also in puppetlabs-firewall, puppetlabs-dism, puppetlabs-vcsrepo

As an alternative, maybe we can do it like e.g. stdlib and stub `defaultprovider`: https://github.com/puppetlabs/puppetlabs-stdlib/blob/08d96808c63535da8280477a01631c9b67463595/spec/unit/facter/service_provider_spec.rb#L13 :
```
Puppet::Type.type(:service).stubs(:defaultprovider).returns provider
```

We talked about how that will impact some type's autorequires, and I'm wondering how provider's even would do that during the compile on the server. A quick review of our supported modules showed only acl does that (https://github.com/puppetlabs/puppetlabs-acl/blob/bec0f1634035785b5f9837b64c9adfe62546e7cb/lib/puppet/type/acl.rb#L394) and I'm very sceptical that this works outside a local `puppet apply` on a windows host, as there would be no access to the AD from a linux puppet master.

lib/rspec-puppet/support.rb
```diff
@@ -217,7 +217,23 @@ def with_vardir
...
+        type.defaultprovider = nil
+      end
+    end
+
     def build_catalog_without_cache(nodename, facts_val, trusted_facts_val, hiera_config_val, code, exported)
+      # Blindly stub out any feature and command Puppet::Confine evaluations
+      # that providers might do. This is necessary because we convert the
+      # resource catalogue to a RAL catalogue in order to evaluate all the
+      # automatic graph edges, which requires evaluating the providers.
+      allow_any_instance_of(Puppet::Confine::Feature).to receive(:pass?).with(anything).and_return(true)
```
DavidS: See https://github.com/puppetlabs/puppetlabs-firewall/blob/cbbd9781164351d38cb8391096b8817e5fa8bf9b/spec/unit/puppet/provider/iptables_spec.rb#L11-L19 for why this is failing on older puppet versions.

lib/rspec-puppet/support.rb
rodjek: > low-level stubs for Puppet::Util::Execution.execute, which will be skipped

Yeah, this stub could be trimmed down so that it's not so encompassing.

> As an alternative, maybe we can do it like e.g. stdlib and stub defaultprovider

It'd need some testing but I fear that this won't be that easy either. Stubbing `Puppet::Type#defaultprovider` with an existing type is easy enough, but what do we pick if the user hasn't specified a provider in their manifest and is expecting Puppet to choose the correct one? We could create a dummy provider, but then we'll need to stub out method in which the providers report their supported features back to the type otherwise we'll run into problems where type parameters become invalid because the provider doesn't support feature X (might not be that big of a problem, but it's not something that I've looked into before so I can't say for sure).

lib/rspec-puppet/support.rb
rodjek: Yeah, it's more a question of "should we even be supporting those versions of Puppet anymore?" :)

DavidS: I'm hesitant to throw away compatibility if all we buy is one less `if` statement. We could alternatively do a 2.99 release similar to how rspec did it, which contains most of the new API and deprecation warnings, and then drop all old shim code when switching over to 3.0 (or 5.0 to match up with puppet's new platform versioning).

A simple workaround would be to guard the whole block here on `Puppet.version > '4'`, and fix it in documentation. It's not like we need to contort ourselves *too* much to keep compatibility.

lib/rspec-puppet/support.rb
```diff
@@ -265,6 +265,20 @@ def build_catalog_without_cache(nodename, facts_val, trusted_facts_val, hiera_co
     end
 
     def stub_facts!(facts)
+      if facts['operatingsystem'] && facts['operatingsystem'] == 'windows'
+        Puppet.settings[:autosign] = false
```
DavidS: Is that required on windows? It could use an explanatory comment if so.

lib/rspec-puppet/support.rb
DavidS: I don't know how provider selection and validation happens on the server. In the end, you're the expert here, and I trust your judgement. In the worst case, we have to revisit this in a few weeks when we get windows testing online.

rodjek:
@DavidS I spent a bit of time this morning looking into the 2.7 support issue and digging further and further down I'm going to make the executive decision to drop 2.7 support. 

Unfortunately, it's not as simple as just adding an if statement for the provider confines. With that problem resolved it exposed another problem in that the same function is used to determine if a string is an absolute path in the provider as in the compiler, so when we stub out `Puppet::Util::Platform.windows?` to allow input validation for the windows file provider to work, it then breaks the ability for the compiler to find manifest files in the filesystem. Given time, I'm sure it could be worked around but I don't think it's worth it just to maintain 2.7 support any more.

DavidS:
Ok, let's get this merged then. For so ancient code, I believe it to be sufficient to note the last "working" version in the README, ala "rspec-puppet 2.5.0 was the last version to support puppet 2.7".

For dropping puppet 3 support, we'll want to do a proper deprecation and version bump cycle.

rodjek:
Closing this out in favour of a new approach that I think will be better.

PR #482, opened 2017-03-16
Stop testing against Puppet 2.x (it's time)

rodjek:
Puppet 3.x has been EOLed, and Puppet 5 is not far away. Let's stop testing against Puppet 2.x now.


DavidS:
There is a workaround for the issue in #481 available, and we don't fully know yet whether we that specific line is really the best solution there.

PR #489, opened 2017-03-31
Automatically filter out defined type being tested from the coverage report

### LINKED ISSUE #159: [Coverage] Testing defines does not remove them from the coverage report
When a define is tested in the rspec tests it doesn't seem to be processed in the coverage reports.

See https://travis-ci.org/elasticsearch/puppet-elasticsearch/jobs/17994337

the python define is tested with the different clients but it still shows up in the untouched resources list.

rodjek:
Previously, this code assumed everything being tested had the same puppet resource type as the example group name, so you'd end up with things like `Node[foo]`, `Class[bar]` being filtered out, but defined types would get a filter like `Define[my::define]` which of course didn't match.

This PR fixes this behaviour so that defined types get a proper filter added to the coverage report.

Closes #159


DavidS:
This conflicts with #490 , and it is not entirely obvious to me how the two changes interact. Please rebase.

Looks good to me otherwise.

PR #305, opened 2015-06-19
Integration 2015 06 19

DavidS:
This contains #223, #228, #246, #296, #300, #301, and #303 with fixes from @javiplx, @tcolgate, @voidus and myself.

PR #494, opened 2017-04-05
Write a full spec_helper on init

### LINKED ISSUE #395: Could not find class
The is the module directory. 

```
➜   tree
.
├── Gemfile
├── Gemfile.lock
├── manifests
│   └── init.pp
├── metadata.json
├── Rakefile
├── README.md
├── spec
│   ├── classes
│   │   └── certs_spec.rb
│   ├── defines
│   ├── fixtures
│   │   ├── manifests
│   │   │   └── site.pp
│   │   └── modules
│   │       └── certs
│   │           └── manifests -> ../../../../manifests
│   ├── functions
│   ├── hosts
│   └── spec_helper.rb
└── tests
    └── init.pp

12 directories, 10 files
```

manifests/init.pp

```
class certs {}
```

spec/classes/certs_spec.rb

```
require 'spec_helper'

describe 'certs' do
  it { is_expected.to compile }
end
```

```
➜   rake rspec
/home/wjiang/.rvm/rubies/ruby-2.1.5/bin/ruby -I/home/wjiang/.rvm/gems/ruby-2.1.5/gems/rspec-core-3.5.1/lib:/home/wjiang/.rvm/gems/ruby-2.1.5/gems/rspec-support-3.5.0/lib /home/wjiang/.rvm/gems/ruby-2.1.5/gems/rspec-core-3.5.1/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
F

Failures:

  1) certs should compile into a catalogue without dependency cycles
     Failure/Error: it { is_expected.to compile }
       error during compilation: Evaluation Error: Error while evaluating a Function Call, Could not find class ::certs for wjiang.engineering.syd2.internal at line 1:1 on node wjiang.engineering.syd2.internal
     # ./spec/classes/certs_spec.rb:4:in `block (2 levels) in <top (required)>'

Finished in 0.10328 seconds (files took 0.53017 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/classes/certs_spec.rb:4 # certs should compile into a catalogue without dependency cycles

/home/wjiang/.rvm/rubies/ruby-2.1.5/bin/ruby -I/home/wjiang/.rvm/gems/ruby-2.1.5/gems/rspec-core-3.5.1/lib:/home/wjiang/.rvm/gems/ruby-2.1.5/gems/rspec-support-3.5.0/lib /home/wjiang/.rvm/gems/ruby-2.1.5/gems/rspec-core-3.5.1/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb failed
```

rodjek:
The change in #300 (merged into master in #305) had the unfortunate side effect of breaking the `fixture_path` calculation which uses `__FILE__` for `spec_helper.rb` as the base for `File.expand_path`. Basically, it would end up configuring rspec-puppet to look for the manifests being tested in the installed gem instead of the user's project directory.

We could work around this by doing clever things like have the spec_helper check the `caller` stack to see if it has been required by another spec_helper and changing the `File.expand_path` base, but in my opinion that feels pretty fragile and I'd prefer to just go back to writing out the file for the user on init.

Fixes #395

/cc @DavidS as this basically undoes his change :)

### DIFF
--- a/lib/rspec-puppet/setup.rb
+++ b/lib/rspec-puppet/setup.rb
@@ -115,7 +115,7 @@ module RSpec::Puppet
     end
 
     def self.safe_create_spec_helper
-      content = "require 'rspec-puppet/spec_helper'\n"
+      content = File.read(File.expand_path(File.join(__FILE__, '..', 'spec_helper.rb')))
       safe_create_file('spec/spec_helper.rb', content)
     end
 
--- a/lib/rspec-puppet/spec_helper.rb
+++ b/lib/rspec-puppet/spec_helper.rb
@@ -5,6 +5,6 @@ fixture_path = File.join(File.dirname(File.expand_path(__FILE__)), 'fixtures')
 RSpec.configure do |c|
   c.module_path     = File.join(fixture_path, 'modules')
   c.manifest_dir    = File.join(fixture_path, 'manifests')
-  c.manifest        = File.join(File.dirname(File.expand_path(__FILE__)), 'fixtures', 'manifests', 'site.pp')
+  c.manifest        = File.join(fixture_path, 'manifests', 'site.pp')
   c.environmentpath = File.join(Dir.pwd, 'spec')
 end
--- a/lib/rspec-puppet/spec_helper.rb
+++ b/lib/rspec-puppet/spec_helper.rb
@@ -1,6 +1,6 @@
 require 'rspec-puppet'
 
-fixture_path = File.expand_path(File.join(__FILE__, '..', 'fixtures'))
+fixture_path = File.join(File.dirname(File.expand_path(__FILE__)), 'fixtures')
 
 RSpec.configure do |c|
   c.module_path     = File.join(fixture_path, 'modules')


DavidS:
:man_facepalming:

I really dislike putting more than a single `require` into the module. It just causes maintenance headaches downstream - especially for people with many modules. I would even go so far as saying that it would be sweet to just add `--require rspec-puppet/spec_helper` and `--require spec_helper` into the `.rspec` file. (see e.g. http://stackoverflow.com/a/5097049). In our internal modulesync_configs it currently looks like this: https://github.com/puppetlabs/modulesync_configs/blob/master/moduleroot/spec/spec_helper.rb#L16 , and I detest that `_local` file. (yes, I know, *I* put this in).

DavidS:
This might be a case where the library and the SDK need to go different paths.

Slightly related, https://github.com/rodjek/rspec-puppet/blob/7d4025ec3c37bb6b2c0d6307451c5872edbe2fe0/lib/rspec-puppet/spec_helper.rb#L8 should also use `fixture_path`, and the original definition of `fixture_path` could use the `dirname` version from L8.

PR #496, opened 2017-04-05
Handle the cases when a resource's name is different to its title

### LINKED ISSUE #448: Resource not found when $name != $title
When testing that a catalogue compiles correctly, we get a failure when the name of a resource is overridden explicitly, and we refer to the new name rather than the title of the resource.

Our specific case is that we use [the Elasticsearch module](https://github.com/elastic/puppet-elasticsearch), which sets `$name` in one of their `service` definitions. See https://github.com/elastic/puppet-elasticsearch/blob/master/manifests/service/init.pp#L196

We use `notify => Service['elasticsearch-cluster']`, where `elasticsearch-cluster` is the value set by `$name`, and not the actual title of the `service` resource. When I run Puppet it works fine, but `rspec-puppet` complains:
```
error during compilation: Could not retrieve dependency 'Service[elasticsearch-support]'
```

rodjek:
As raised in #448, rspec-puppet doesn't properly handle resources where the title is different to the name.

This PR fixes the behaviour with a number of changes:
 * When using the `compile` matcher, an entry is added to the resource hash using the namevar (in addition to the existing entries for the aliases and title)
 * When calculating resource relationships, add references to the resources by the name (in addition to the existing references by title) to the results.
 * When compiling the catalogue, an alias is created in the catalogue so that the resource can be referred to by its name as well as its title

h/t to @domcleal for providing the test cases

Closes #448

### DIFF
--- a/spec/classes/relationship__titles_spec.rb
+++ b/spec/classes/relationship__titles_spec.rb
@@ -1,0 +1,14 @@
+require 'spec_helper'
+
+describe 'relationships::titles' do
+  it { should compile }
+  it { should compile.with_all_deps }
+
+  it { should contain_file('/etc/svc') }
+  it { should contain_service('svc-title') }
+
+  it { should contain_file('/etc/svc').that_notifies('Service[svc-name]') }
+  it { should contain_file('/etc/svc').that_comes_before('Service[svc-name]') }
+  it { should contain_service('svc-title').that_requires('File[/etc/svc]') }
+  it { should contain_service('svc-title').that_subscribes_to('File[/etc/svc]') }
+end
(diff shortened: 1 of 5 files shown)


spec/classes/relationship__titles_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'spec_helper'
+
+describe 'relationships::titles' do
+  let(:facts) { {:operatingsystem => 'Debian', :kernel => 'Linux'} }
+
+  it { should compile }
+  it { should compile.with_all_deps }
+
+  it { should contain_file('/etc/svc') }
+  it { should contain_service('svc-title') }
```
DavidS: should this also test for `should contain_service('svc-name')` ?

Issue #192, opened 2014-04-27
compile matcher fails when checking classes targeting windows on posix.


fessyfoo:
We're trying to  use the compile matcher on all of our role classes. Our puppetmaster serves windows and linux nodes.   each node definition has one role class assigned.

when trying to compile a windows role, on a ci linux or dev osx machine.  there's inevitably a File resource with a path of the form: `C:\foo.txt`   rspec-puppet fails to compile the class because `File paths must be fully qualified, not 'C:\foo.txt'`  

Any guidance on how to resolve this would be greatly appreciated. 

for those interested I've created a sample project which illustrates the issue:

   https://github.com/fessyfoo/rspec-puppet-windows-issue

fessyfoo:
Just reiterating that the puppetmaster running on linux  does not have a problem compiling catalogs for windows nodes,  so it's very odd that the specs do.    I got lost in the puppet code for this but it almost looked like the puppetmaster and the spec code were calling different catalog complilation paths within puppet.   ...   fuzzy I know.

philippvh:
+1: Same issue happening here.
I tried to apply work-around from @drrb. It works to a point, but there seem to be more dependencies in our case than in the example code. I won't spend time on it currently, just as a note here.

mrdima:
+1 for fixing this

mbruntink:
+1, same issue here.

DavidS:
I've delved deeper into this issue. `puppet master --compile` creates only a so-called 'resource catalog', which is the server-side representation to facilitate network serialization. This does not call into any of the validation methods nor creates a "proper" resource graph.

rspec-puppet on the other hand, needs the resource graph and therefore acts like an agent and converts the resource catalog in a RAL catalog. This starts calling into actual types and therefore hits this validation.

This is documented in https://github.com/puppetlabs/puppet/blob/master/docs/catalogs.md

Stubbing `Puppet::Util::Platform.windows?`, as @drrb can only be a start in this case. On the puppet side alone, there is also `Puppet.features.microsoft_windows?` which needs to be considered. Finally, all types that are used by the catalog need to be checked for things where they access system specific runtime behaviour.

In the end it might be easier to run rspec on an actual windows system for these cases. rspec's metadata support might help here to select tests based on platform.

fessyfoo:
thanks for that explanation.

mcasper:
Would love to see this fixed

[...]

PR #498, opened 2017-04-10
(MODULES-4287) Make rspec-puppet tests platform independent

### LINKED ISSUE #175: 'should compile' doesn't like 'groups' in 'user' definitions

rodjek:
In order to fully resolve the graph and create all the automatic relationships we have to convert the resource catalogue produced by the compiler into a RAL catalogue like that which an agent would apply.

An unfortunate side effect of this is that it causes providers to be loaded and tested for suitability. This can cause problems for people who have manifests that specify providers for platforms other than the one they're running rspec-puppet on (e.g. testing Windows manifests on a Linux host).

This PR adds a few monkey patches make this possible:

 * `Puppet::Type#set_default` has been modified so that we can prevent Puppet from assigning a default provider when performing catalogue tests (but not when testing the types themselves).
 * `Puppet::Util::Platform.windows?` has been modified so that can force Puppet to pretend that it's running on windows, allowing the windows specific parameter validations to succeed.
 * `Kernel.require` has been modified to prevent Puppet from requiring `puppet/util/windows` if Puppet is pretending to be running on windows in order to prevent it from requiring other libraries that generally aren't installed on non-windows hosts.
 * `Puppet::Node::Environment#validate_dirs` (if it exists, which it only does on Puppet < 3.x) has been wrapped to disable the windows pretending (if we're pretending) while the method is running as this is used by the compiler to determine if the paths on disk to the manifests is valid.

This PR supercedes #481

Closes #175 #192 #256 #352 #376 #387 #437

### DIFF
--- a/lib/rspec-puppet/monkey_patches.rb
+++ b/lib/rspec-puppet/monkey_patches.rb
@@ -9,8 +9,32 @@ module Puppet
     old_set_default = instance_method(:set_default)
 
     define_method(:set_default) do |attr|
-      return if attr == :provider && self.class.suppress_provider?
-      old_set_default.bind(self).call(attr)
+      old_posix = nil
+      old_microsoft_windows = nil
+
+      if attr == :provider
+        old_posix = Puppet.features.posix?
+        old_microsoft_windows = Puppet.features.microsoft_windows?
+
+        if Puppet::Util::Platform.pretend_windows?
+          Puppet.features.add(:posix) { false }
+          Puppet.features.add(:microsoft_windows) { true }
+        else
+          Puppet.features.add(:posix) { true }
+          Puppet.features.add(:microsoft_windows) { false }
+        end
+      end
+
+      retval = old_set_default.bind(self).call(attr)
+
+      unless old_posix.nil?
+        Puppet.features.add(:posix) { old_posix }
+      end
+      unless old_microsoft_windows.nil?
+        Puppet.features.add(:microsoft_windows) { old_microsoft_windows }
+      end
+
+      retval
     end
 
     def self.suppress_provider?
@@ -37,16 +61,15 @@ module Puppet
       old_validate_dirs = instance_method(:validate_dirs)
 
       define_method(:validate_dirs) do |dirs|
-        pretending = false
+        pretending = Puppet::Util::Platform.pretend_platform
 
-        if Puppet::Util::Platform.pretend_windows?
-          pretending = true
-          Puppet::Util::Platform.unpretend_windows
+        if pretending
+          Puppet::Util::Platform.pretend_to_be nil
         end
 
         output = old_validate_dirs.bind(self).call(dirs)
 
-        Puppet::Util::Platform.pretend_windows if pretending
+        Puppet::Util::Platform.pretend_to_be pretending
 
         output
       end
@@ -54,30 +77,48 @@ module Puppet
   end
 
   module Util
-    # Allow rspec-puppet to pretend to be windows.
+    # Allow rspec-puppet to pretend to be different platforms.
     module Platform
       def windows?
-        pretend_windows? || !!File::ALT_SEPARATOR
+        pretend_platform.nil? ? (actual_platform == :windows) : pretend_windows?
       end
       module_function :windows?
 
+      def actual_platform
+        @actual_platform ||= !!File::ALT_SEPARATOR ? :windows : :posix
+      end
+      module_function :actual_platform
+
       def pretend_windows?
-        @pretend_windows ||= false
+        pretend_platform == :windows
       end
       module_function :pretend_windows?
 
-      def pretend_windows
-        @pretend_windows = true
-      end
-      module_function :pretend_windows
+      def pretend_to_be(platform)
+        # Ensure that we cache the real platform before pretending to be
+        # a different one
+        actual_platform
 
-      def unpretend_windows
-        @pretend_windows = false
+        @pretend_platform = platform
       end
-      module_function :unpretend_windows
+      module_function :pretend_to_be
+
+      def pretend_platform
+        @pretend_platform ||= nil
+      end
+      module_function :pretend_platform
     end
   end
 end
+
+# Prevent the File type from munging paths (which uses File.expand_path to
+# normalise paths, which does very bad things to *nix paths on Windows.
+Puppet::Type.type(:file).paramclass(:path).munge { |value| value }
+
+# Prevent the Exec type from validating the user. This parameter isn't
+# supported under Windows at all and only under *nix when the current user is
+# root.
+Puppet::Type.type(:exec).paramclass(:user).validate { |value| true }
 
 # Prevent Puppet from requiring 'puppet/util/windows' if we're pretending to be
 # windows, otherwise it will require other libraries that probably won't be
--- a/lib/rspec-puppet/monkey_patches.rb
+++ b/lib/rspec-puppet/monkey_patches.rb
@@ -1,0 +1,91 @@
+module Puppet
+  # Allow rspec-puppet to prevent Puppet::Type from automatically picking
+  # a provider for a resource. We need to do this because in order to fully
+  # resolve the graph edges, we have to convert the Puppet::Resource objects
+  # into Puppet::Type objects so that their autorequires are evaluated. We need
+  # to prevent provider code from being called during this process as it's very
+  # platform specific.
+  class Type
+    old_set_default = instance_method(:set_default)
+
+    define_method(:set_default) do |attr|
+      return if attr == :provider && self.class.suppress_provider?
+      old_set_default.bind(self).call(attr)
+    end
+
+    def self.suppress_provider?
+      @suppress_provider ||= false
+    end
+
+    def self.suppress_provider
+      @suppress_provider = true
+    end
+
+    def self.unsuppress_provider
+      @suppress_provider = false
+    end
+  end
+
+  # If Puppet::Node::Environment has a validate_dirs instance method (i.e.
+  # Puppet < 3.x), wrap the method to check if rspec-puppet is pretending to be
+  # running under windows. The original method uses Puppet::Util.absolute_path?
+  # (which in turn calls Puppet::Util::Platform.windows?) to validate the path
+  # to the manifests on disk during compilation, so we have to temporarily
+  # disable the pretending when running it.
+  class Node::Environment
+    if instance_methods.include?("validate_dirs")
+      old_validate_dirs = instance_method(:validate_dirs)
+
+      define_method(:validate_dirs) do |dirs|
+        pretending = false
+
+        if Puppet::Util::Platform.pretend_windows?
+          pretending = true
+          Puppet::Util::Platform.unpretend_windows
+        end
+
+        output = old_validate_dirs.bind(self).call(dirs)
+
+        Puppet::Util::Platform.pretend_windows if pretending
+
+        output
+      end
+    end
+  end
+
+  module Util
+    # Allow rspec-puppet to pretend to be windows.
+    module Platform
+      def windows?
+        pretend_windows? || !!File::ALT_SEPARATOR
+      end
+      module_function :windows?
+
+      def pretend_windows?
+        @pretend_windows ||= false
+      end
+      module_function :pretend_windows?
+
+      def pretend_windows
+        @pretend_windows = true
+      end
+      module_function :pretend_windows
+
+      def unpretend_windows
+        @pretend_windows = false
+      end
+      module_function :unpretend_windows
+    end
+  end
+end
+
+# Prevent Puppet from requiring 'puppet/util/windows' if we're pretending to be
+# windows, otherwise it will require other libraries that probably won't be
+# available on non-windows hosts.
+module Kernel
+  alias :old_require :require
+  def require(path)
+    return if path == 'puppet/util/windows' && Puppet::Util::Platform.pretend_windows?
+    old_require(path)
+  end
+end
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -272,6 +272,19 @@ module RSpec::Puppet
     end
 
     def stub_facts!(facts)
+      if facts['operatingsystem'] && facts['operatingsystem'].to_s.downcase == 'windows'
+        # If we're not running in windows but are testing a catalogue destined
+        # for a windows host, pretend to be windows in the least awful way that
+        # I can think of.
+        unless Puppet::Util::Platform.windows?
+          Puppet::Util::Platform.pretend_windows
+          # On windows, puppet validates the value of the autosign setting as
+          # an absolute path unless it's disabled.
+          Puppet.settings[:autosign] = false
+        end
+      else
+        Puppet::Util::Platform.unpretend_windows
+      end
       facts.each { |k, v| Facter.add(k) { setcode { v } } }
     end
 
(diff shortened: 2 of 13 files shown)


lib/rspec-puppet/monkey_patches.rb
```diff
@@ -0,0 +1,91 @@
...
+    def self.suppress_provider
+      @suppress_provider = true
+    end
+
+    def self.unsuppress_provider
+      @suppress_provider = false
+    end
+  end
+
+  # If Puppet::Node::Environment has a validet_dirs instance method (i.e.
```
james-stocks: Small typo on validet_dirs

lib/rspec-puppet/support.rb
```diff
@@ -254,6 +254,19 @@ def build_catalog_without_cache(nodename, facts_val, trusted_facts_val, hiera_co
...
     def stub_facts!(facts)
+      if facts['operatingsystem'] && facts['operatingsystem'].to_s.downcase == 'windows'
+        # If we're not running in windows but are testing a catalogue destined
+        # for a windows host, pretend to be windows in the least awful way that
+        # I can think of.
+        unless Puppet::Util::Platform.windows?
+          Puppet::Util::Platform.pretend_windows
+          # On windows, puppet validates the value of the autosign setting as
+          # an absolute path unless it's disabled.
+          Puppet.settings[:autosign] = false
```
james-stocks: Is there a risk of breaking spec tests; if someone happened to be asserting this was true? I think if so; we could put somewhere in docs that we have this restriction when testing a module with Windows facts

lib/rspec-puppet/monkey_patches.rb
rodjek: Thanks, fixed :)

lib/rspec-puppet/support.rb
rodjek: I don't think it's possible to break any specs by enforcing this value. It's a master setting and doesn't have any effect on the compilation of the catalogue.

lib/rspec-puppet/support.rb
```diff
@@ -15,6 +15,27 @@ def environment
...
 
+    def stub_file_consts(example)
+      munged_facts = facts_hash(nodename(example.metadata[:type]))
+
+      if munged_facts['operatingsystem'] && munged_facts['operatingsystem'].to_s.downcase == 'windows'
+        stub_const_wrapper('PATH_SEPARATOR', ';')
+        stub_const_wrapper('ALT_SEPARATOR', "\\")
+      else
+        stub_const_wrapper('PATH_SEPARATOR', ':')
+        stub_const_wrapper('ALT_SEPARATOR', nil)
```
DavidS: Seeing those values hardcoded here makes me very itchy. Please add a set of tests to verify the underlying ruby platform like this:

```
# make sure we do not miss changes in the ruby platform relevant for lib/rspec-puppet/support.rb stub_file_consts
describe "consts on windows", :if => RSpec::Support::OS.windows? do
  it { expect(File.PATH_SEPARATOR).to eq(':') }
  it { expect(File.ALT_SEPARATOR).to be_nil }
end

describe "consts on linux", :unless => RSpec::Support::OS.windows? do
  it { expect(File.PATH_SEPARATOR).to eq(';') }
  it { expect(File.ALT_SEPARATOR).to eq("\\") }
end
```

lib/rspec-puppet/support.rb
```diff
@@ -254,6 +275,12 @@ def build_catalog_without_cache(nodename, facts_val, trusted_facts_val, hiera_co
     end
 
     def stub_facts!(facts)
+      if facts['operatingsystem'] && facts['operatingsystem'].to_s.downcase == 'windows'
+        Puppet::Util::Platform.pretend_to_be :windows
+      else
+        Puppet::Util::Platform.pretend_to_be :linux
+      end
+      Puppet.settings[:autosign] = false
```
DavidS: Preserve the note on Windows validating the value here, and note that the setting is not used during unit-testing anyways.

lib/rspec-puppet/monkey_patches.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+end
+
+# Prevent Puppet from requiring 'puppet/util/windows' if we're pretending to be
+# windows, otherwise it will require other libraries that probably won't be
+# available on non-windows hosts.
+module Kernel
+  alias :old_require :require
+  def require(path)
+    return if path == 'puppet/util/windows' && Puppet::Util::Platform.pretend_windows?
```
DavidS: Will this be confusing/dangerous to people? Do we have a (easy) way to emit a single debug/warning message when this happens?

lib/rspec-puppet/support.rb
rodjek: Yeah, it's a nasty solution for a nasty problem (Puppet really doesn't want to do what we're forcing it to do) 😀

james-stocks:
I've made a rough attempt to run these changes past appveyor (rough as in the appveyor config might not be totally correct and appropriate yet) and am seeing 7 failures: https://ci.appveyor.com/project/rodjek/rspec-puppet/build/1.0.12/job/f7rgrritnbw4y2i8

```
rspec ./spec/classes/relationship__before_spec.rb:22 # relationships::before should contain File[/tmp/foo] that comes before File[/tmp/foo/bar]
rspec ./spec/classes/relationship__before_spec.rb:23 # relationships::before should contain File[/tmp/foo/bar] that requires File[/tmp/foo]
rspec ./spec/classes/relationship__before_spec.rb:25 # relationships::before should contain Notify[bazz] that comes before File[/tmp/foo/bar]
rspec ./spec/classes/relationship__before_spec.rb:26 # relationships::before should contain Notify[qux] that requires File[/tmp/foo]
rspec ./spec/classes/relationship__before_spec.rb:27 # relationships::before should contain Notify[bazz] that comes before Notify[qux]
rspec ./spec/classes/relationship__before_spec.rb:28 # relationships::before should contain Notify[qux] that requires Notify[bazz]
rspec ./spec/classes/test_duplicate_alias_spec.rb:4 # test::duplicate_alias should compile into a catalogue without dependency cycles
```

rodjek:
OK, I think this is finally ready to go. rspec-puppet's tests now pass (at least on my windows 10 vm) which should unblock #505.

On my linux workstation, I can run tests for windows modules (puppetlabs-wsus_client, puppetlabs-acl, puppetlabs-chocolatey were used for testing). On my windows 10 vm, I can run tests for linux modules (puppetlabs-ntp, puppetlabs-apt, puppetlabs-apache, puppetlabs-puppetdb were used for testing).

This is very much "best effort" support due to Puppet not being designed to do what we're trying to make it do (particularly in the case of compiling catalogues for non-Windows agents on Windows). All the core types should work fine, but 3rd party types might still have issues.

Iristyle:
In particular https://github.com/puppetlabs/puppetlabs-acl/blob/master/spec/unit/type/acl_spec.rb#L277-L279 is now broken as a result of the monkey patches to Puppets file type.  There may be more fallout.

I'd suggest rethinking that change.

DavidS:
Hi @Iristyle,

I created a ticket for you tracking this.

Cheers, David

cdenneen:
@rodjek @DavidS looks like this is broken in 2.7.0
reverted back to 2.6.15 and working again.

rodjek:
@cdenneen can you send me a simple example module that shows what's broken?

franknarf8:
I seem to have an exec that fails because of the `creates` argument
Running the code on Windows; for MacOS.
Complains with : `error during compilation: Parameter creates failed on Exec[installs something]: creates must be a fully qualified path`
 ```
exec { "installs something":
    command     => "whatever",
    path        => '/usr/local/bin:/usr/bin:/bin',
    environment => ["HOME=/Users/myuser"],
    user        => 'myuser',
    creates     => "/Users/myuser/some/further/path"
}
```

**Comment by rodjek** [comment]
@franknarf8 Can you open a new issue for this please and include the contents of the spec file as well (these sort of issues are often caused by not setting the correct facts in the tests)

**Comment by cdenneen** [comment]
Using 2.7.5

```
  1) profile::webfeeds should compile into a catalogue without dependency cycles
     Failure/Error: it { is_expected.to compile }
       error during compilation: Parameter path failed on File[c:\temp]: File paths must be fully qualified, not 'c:\temp' (file: /Users/cdenneen/src/gitlab/keystone-puppet/spec/fixtures/modules/staging/manifests/init.pp, line: 26)
```

This is set in the profile::webfeeds as:

```
class profile::webfeeds(
  $temp                          = 'c:\\temp',
```

So the `\\` doesn't seem to fix it here.

@rodjek @DavidS

uberjew666:
I'm also getting error:
`./spec/classes/profile/os/directories_spec.rb:27: error during compilation: Parameter path failed on File[C:/Scripts/]: File paths must be fully qualified, not 'C:/Scripts'`

Not sure which version, I'm using PDK 1.10

PR #500, opened 2017-04-11
Handle resources with identical namevars but unique titles

### LINKED ISSUE #499: Error raised for resources with identical aliases, but different titles
a3d3cbc (I think) introduced a regression building catalogs which contain multiple resources with different titles, but identical aliases.

    exec { 'foo_bar_1':
      command => '/bin/echo foo bar',
    }
    exec { 'foo_bar_2':
      command => '/bin/echo foo bar',
    }

rspec-puppet would add identical aliases to the catalog for the two execs, causing a duplicate and the following error:

     ArgumentError:
       Cannot alias Exec[foo_bar_2] to "/bin/echo foo bar" at spec/fixtures/modules/relationships/manifests/titles.pp:15; resource ["Exec", "/bin/echo foo bar"] already declared at spec/fixtures/modules/relationships/manifests/titles.pp:12
     # gems/puppet-4.9.4/lib/puppet/resource/catalog.rb:190:in `alias'
     # ./lib/rspec-puppet/support.rb:256:in `block in build_catalog_without_cache'
     # ./lib/rspec-puppet/support.rb:254:in `each'
     # ./lib/rspec-puppet/support.rb:254:in `build_catalog_without_cache'
     # ./lib/rspec-puppet/support.rb:268:in `block in build_catalog'
     # ./lib/rspec-puppet/cache.rb:17:in `call'
     # ./lib/rspec-puppet/cache.rb:17:in `get'
     # ./lib/rspec-puppet/support.rb:267:in `build_catalog'
     # ./lib/rspec-puppet/support.rb:31:in `block in load_catalogue'
     # ./lib/rspec-puppet/support.rb:219:in `with_vardir'
     # ./lib/rspec-puppet/support.rb:19:in `load_catalogue'
     # ./lib/rspec-puppet/example/class_example_group.rb:7:in `catalogue'
     # ./lib/rspec-puppet/support.rb:11:in `block in subject'
     # ./lib/rspec-puppet/matchers/compile.rb:23:in `call'
     # ./lib/rspec-puppet/matchers/compile.rb:23:in `matches?'
     # ./spec/classes/relationship__titles_spec.rb:6:in `block (2 levels) in <top (required)>'

This exec pattern is used in puppetlabs-postgresql here: https://github.com/puppetlabs/puppetlabs-postgresql/blob/4.9.0/manifests/server/config_entry.pp#L48-L64

Example test: https://gist.github.com/domcleal/bbee33af2d1626fa99cda972bc2dfe8f

rodjek:
As raised in #499, the change introduced in #496 broke support for cases where resources of the same type have the same value in their namevar parameter but have unique titles to prevent duplicate resource declaration. This was due to the way we used resource aliases in the catalogue to handle the issue of resources where title != namevar.

This PR changes the behaviour to remove the resource aliases and instead change the resource lookup logic to fallback to iterating through all the resources in the catalogue of the requested type and checking their namevars.

I've also removed the test from the `relationships::titles` class where we were allowing users to do the initial search for the resource by namevar instead of title (functionality that was only introduced in #496 and not part of any release). In my opinion, you should test for the presence of the resource (`contain_service('foo')`) by the title you gave it in the manifest. If there are strong feelings against this though, it wouldn't be hard to reimplement.

Closes #499

### DIFF
--- a/spec/classes/test_duplicate_alias_spec.rb
+++ b/spec/classes/test_duplicate_alias_spec.rb
@@ -1,0 +1,5 @@
+require 'spec_helper'
+
+describe 'test::duplicate_alias' do
+  it { should compile }
+end
--- a/spec/classes/test_duplicate_alias_spec.rb
+++ b/spec/classes/test_duplicate_alias_spec.rb
@@ -2,4 +2,7 @@ require 'spec_helper'
 
 describe 'test::duplicate_alias' do
   it { should compile }
+  it { should contain_exec('foo_bar_1') }
+  it { should contain_exec('foo_bar_2') }
+  it { should_not contain_exec('/bin/echo foo bar') }
 end
--- a/lib/rspec-puppet/matchers/create_generic.rb
+++ b/lib/rspec-puppet/matchers/create_generic.rb
@@ -214,7 +214,7 @@ module RSpec::Puppet
 
       def check_befores(catalogue, resource)
         @befores.each do |ref|
-          unless precedes?(resource, catalogue.resource(ref))
+          unless precedes?(resource, canonicalize_resource(ref))
             @errors << BeforeRelationshipError.new(resource.to_ref, ref)
           end
         end
@@ -222,7 +222,7 @@ module RSpec::Puppet
 
       def check_requires(catalogue, resource)
         @requires.each do |ref|
-          unless precedes?(catalogue.resource(ref), resource)
+          unless precedes?(canonicalize_resource(ref), resource)
             @errors << RequireRelationshipError.new(resource.to_ref, ref)
           end
         end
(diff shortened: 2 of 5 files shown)


spec/classes/test_duplicate_alias_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
+
+describe 'test::duplicate_alias' do
+  it { should compile }
```
DavidS: should you add positive and negative tests here that `foo_bar_1`, and `foo_bar_2` can be found, but not `/bin/echo foo bar`?

rodjek: Good idea, can never have too many tests! Added :)

DavidS: The things I have seen say otherwise. #endisnigh

PR #507, opened 2017-04-26
Adjust facts handling to only downcase fact names in facts hash

seanmil:
Only the top-level keys, corresponding to the fact names, in the
facts hash should be downcased. Puppet leaves any keys as part
of a hash returned from a structured fact in the original case.


seanmil:
Is there anything needed on this PR? Just hoping it doesn't get missed if there is a new release in the works.

DavidS:
Thanks for your contribution. No worries, we would have done a triage/merge pass before release anyways.

PR #517, opened 2017-05-17
Added a Post Condition Hook

### LINKED ISSUE #516: Need a post_condition hook
I've run into cases a couple of times where I needed a `post_condition` in addition to the existing `pre_condition` to do what I need to do.

I get that this is usually indicative of a code-smell, but, when messing about with the guts of Puppet, you may need to order things around what you're testing.

trevor-vaughan:
This adds a 'post_condition' hook that compliments the 'pre_condition'
hook but allows you to place code *after* the current code.

Closes #516

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,6 +502,23 @@ relies on a top-level `apache` class being included first.
 The value may be a raw string to be inserted into the Puppet manifest, or an array of strings
 (manifest fragments) that will be concatenated.
 
+#### Specifying extra code to load (post-conditions)
+
+In some cases, you may need to ensure that the code that you are testing comes
+**before** another set of code. Similar to the `:pre_condition` hook, you can add
+a `:post_condition` hook that will ensure that the added code is evaluated
+**after** the tested class.
+
+```ruby
+let(:post_condition) { 'include other_class' }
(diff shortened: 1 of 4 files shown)


trevor-vaughan:
@rodjek Any changes needed for this?

DavidS:
The code itself seems to be solid! 

Please add a short description of `post_condition` to the README at https://github.com/rodjek/rspec-puppet/#specifying-extra-code-to-load-pre-conditions.

It also seems that the tests should verify that the file has made its way into the catalog.

trevor-vaughan:
@DavidS Updated the README. I have no idea how to test that the catalog is correct since `pre_condition` wasn't tested in the original test suite. It certainly works in anger :-D.

DavidS:
@trevor-vaughan added the tests at the "right" place. If you concur, we can squash the whole branch down to a single commit for merging.

trevor-vaughan:
@DavidS Works for me. Squash and Merge away (and thanks for adding the tests)!

DavidS:
I had to update it a bit to make it work for puppet 3 and earlier, but now it passes.

PR #519, opened 2017-05-23
Support setting node parameters

tdevelioglu:
This commit adds a support for setting a node's parameters. It adds a
new helper, :node_params, that is a hash with members that are passed to
the instantiated node object to become its top-scope parameters.

Currently the only ways to set top scope parameters in an example is to
specify them as facts (which isn't technically really correct and also sets
the same values in the facts hash) or set them as a :pre_condition which,
as a string, is a pain to work with when you'd like to override things
in individual examples.

Adding support for :node_params, makes it possible to set top-scope
variables for a node the same way you would when using an ENC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -487,6 +487,29 @@ end
 Any facts you provide with `let(:facts)` in a spec will automatically be merged on top
 of the default facts.
 
+#### Specifying top-scope variables that should be available to your manifest
+
+You can create top-scope variables much in the same way as an ENC.
+
+
+```ruby
+let(:node_params) { { :hostgroup => 'webservers', :rack => 'KK04', :status => 'maintenance' } }
+```
+
+You can also create a set of default top-scope variables provided to all specs in your spec_helper:
+
+``` ruby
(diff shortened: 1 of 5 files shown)


tdevelioglu:
CI tests for puppet 2.7 are failing, can we ignore those ? From outstanding PR's it looks like puppet 2.x support is being removed anyway.

DavidS:
You can shield your tests from executing on 2.7 using something like here: https://github.com/rodjek/rspec-puppet/blob/master/spec/applications/orch_app_spec.rb#L3 

When explaining your feature in the README, please also add a note on the version range supported.

tdevelioglu:
@DavidS Done!

DavidS:
I've added another test to gain confidence that this change does not impact facts. I'll merge it when travis comes back green.

This should be squashed into a single commit on merge.

DavidS:
updated to avoid puppet 3, which doesn't have a facts hash yet (and therefore ca never run into this problem)

PR #528, opened 2017-06-11
Experimental Pathname#absolute? stubbing

rodjek:
Definitely not something I'd want enabled by default.

### DIFF
--- a/lib/rspec-puppet.rb
+++ b/lib/rspec-puppet.rb
@@ -14,6 +14,10 @@ begin
 rescue LoadError
 end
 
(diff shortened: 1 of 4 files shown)


DavidS:
Definitely!

Could you instead add a shared context adding stubs:

```
RSpec.shared_context "with Pathname.absolute stubbed"
  before(:each) do
   allow(Pathname).to receive(:absolute?) do |arg|
     # ...
```

This would also be beneficial, by leaving the original object unpatched in the default case, and leaving the patching up to a dedicated library.

PR #540, opened 2017-07-02
Fix race condition on mkdir and symlink under parallel tests

kpaulisse:
This PR fixes a race condition that we experienced because we run rspec-puppet tests parallelized. Without this fix we were periodically running into cases where this would happen:

- process 1 would notice that a directory is missing
- process 2 would notice that a directory is missing
- process 1 would create the directory ✅ 
- process 2 would create the directory and get Errno::EEXIST 🚫

The change here disregards the `Errno::EEXIST` for directories and symlinks, when the directory or symlink that's being created already exists and is proper.

### DIFF
--- a/lib/rspec-puppet/setup.rb
+++ b/lib/rspec-puppet/setup.rb
@@ -97,7 +97,11 @@ module RSpec::Puppet
           $stderr.puts "!! #{dir} already exists and is not a directory"
         end
       else
-        FileUtils.mkdir dir
+        begin
+          FileUtils.mkdir dir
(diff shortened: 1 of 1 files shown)


rodjek:
Thanks for the patch! I'll see if there are any other issues in 2.6.1 that come in today and then push out a 2.6.2 release with this tonight.

kpaulisse:
Cool thanks. Hope you're doing well.

rodjek:
rspec-puppet 2.6.2 pushed to rubygems

PR #541, opened 2017-07-03
Add 'setup_fixtures' configuration option

scottatron:
When using a modulepath that already satisfies the correct structure, the call to `RSpec::Puppet::Setup.safe_setup_directories` emits the 'Unable to determine module name.' error message despite the specs continuing to run correctly.

This adds a configuration option to disable the default fixtures setup.


rodjek:
rspec-puppet 2.6.2 pushed to rubygems

Issue #549, opened 2017-07-03
2.6 is pulling the `fqdn` fact from the base OS


trevor-vaughan:
Since upgrading to the `2.6` release of `rspec-puppet` our tests are now pulling from the base host's `fqdn` (not sure what else at this point) instead of using the `let(:facts)` version of the values.

This is presently breaking most of our test runs.

rnelson0:
Given an rspec file like this:
```puppet
  on_supported_os.each do |os, facts|
    next unless facts[:kernel] == 'Linux'
    context "on #{os} " do
      let :facts do
        facts.merge({
          clientcert: 'wiki',
        })
      end

      context 'when using default params' do
        let :params do
          params
        end
```
And a hiera config in `spec/spec_helper.rb` like this:
```
RSpec.configure do |c|
  c.hiera_config = File.expand_path(File.join(__FILE__, '../fixtures/hiera.yaml'))
  default_facts = {
    puppetversion: Puppet.version,
    facterversion: Facter.version
  }
  default_facts.merge!(YAML.load(File.read(File.expand_path('../default_facts.yml', __FILE__)))) if File.exist?(File.expand_path('../default_facts.yml', __FILE__))
  c.default_facts = default_facts
end
```
And a hiera.yaml like this:
```
---
:backends:
  - yaml
:yaml:
  :datadir: './spec/fixtures/hieradata'
:hierarchy:
  - '%{::clientcert}'
  - 'default'
```
And a validation in the class like this:
```puppet
  if($packages == {}) {
    fail('You must provide a hash of packages for wiki implementation.')
  }
  create_resources('package', $packages)
```
Then, I expect class parameters to be looked up from the file `./spec/fixtures/hieradata/wiki.yaml' and my tests to pass, as they do with rspec-puppet <2.6.0. Instead, with rspec-puppet >= 2.6.0, I receive errors like this:
```
     Puppet::PreformattedError:
       Evaluation Error: Error while evaluating a Function Call, You must provide a hash of packages for implementation. at /.../spec/fixtures/modules/X/manifests/init.pp:109:5 on node Y
```
I have pinned my gemfile to rspec-puppet 2.5.x for now: `gem "rspec-puppet", '~> 2.5.0'`

trevor-vaughan:
@rodjek Requesting that 2.6.0 be pulled unless this is an easy fix.

ekohl:
[...]

PR #550, opened 2017-07-04
Merge node derived facts only if the node name has been specified

rodjek:
Re #549 

@trevor-vaughan, @rnelson0, @ekohl - I believe this should solve your issue, so if you could run it against your affected modules and let me know, that'd be great.

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -228,11 +228,21 @@ module RSpec::Puppet
                        {}
                      end
 
+      # Merge in node facts so they always exist by default, but only if they
+      # haven't been defined in `RSpec.configuration.default_facts`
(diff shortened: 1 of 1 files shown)


rnelson0:
I can confirm that this works for my code (a controlrepo, not a component module).

rodjek:
Thanks @rnelson0! I'll wait for the others to respond before merging this in, but in the meantime I'll work up some tests so that we don't see this regression in the future.

rodjek:
@trevor-vaughan confirmed this fixes the issue(s) he was seeing in #549 as well.

alexjfisher:
Fixes test failures seen in https://github.com/voxpupuli/puppet-rundeck/pull/322

mmoll:
Fixes the problems we're seeing at @theforeman modules 👍

rodjek:
Thanks all, 2.6.3 has just been pushed to rubygems

Issue #565, opened 2017-07-11
No longer able to explicitly set hostname, fqdn or domain facts


seanmil:
With the changes as a result of the report starting in, I think, #509 it is no longer possible to set a node name independent from the system's facter FQDN. This is a breaking change from the prior release and has broken a number of our test cases.

In environments where the node (certificate) name of a system is independent from its FQDN this introduces significant testing issues as there is no longer any way to set the fqdn, hostname, or domain facts (either top-level or in the networking hash) to the expected/required values for testing.

Previously in facts_hash https://github.com/rodjek/rspec-puppet/blob/c6639523b7989fd1c935905f0348f4267d288442/lib/rspec-puppet/support.rb#L113 any user-specified facts were merged in last.

seanmil:
In order to stick with semver, I think any solution will require letting people override the facts via setting them in the facts hash.

A way to support this going forward might be to let people do something like: `let(:fqdn) { 'test.example.com' }` to set the value that the `fqdn`, `domain`, and `hostname` facts derive from, and if no `let(:fqdn)` is declared then default to the value of `:node`.

seanmil:
I've just pushed a fix which I believe addresses both the original concerns/goals as well as provides compatibility with user-supplied fact overrides.

rodjek:
Yeah, I've never actually seen a real world use of nodename != fqdn, it's always been a theoretical case until now! :)

Given that, I don't think it's possible to satisfy all the conditions without a breaking change. I think I'll have to revert the original change and restore the 2.5.x behavior. I like the idea of using `let(:fqdn)`, but that's still a breaking behavior change. As it would require a major version bump, I'd also like to rename `let(:node)` to something like `let(:certname)` to make the difference between the two obvious.

seanmil:
Having nodename != fqdn is a requirement in our environment for a couple of reasons. It does make some things a bit more complicated in Puppet-land, given the general presumption that they are equal, but it is quite doable.

Did you have a chance to look at the PR I put up? Any reason to think it wouldn't work to achieve the original intention that prompted the change while still providing backwards compatibility?

rodjek:
Yeah, the change that was introduced in 2.6.0 to override the fqdn based facts was to account for the rspec-puppet-facts workflow. rspec-puppet-facts gives you a hash of pre-generated facts for a particular OS, which includes things like `fqdn`, `hostname`, `domain` etc and we need to be able to override those values if the user specifies a node name in their spec.

[...]

PR #566, opened 2017-07-11
Fix regression in fact fqdn/domain/host override behavior

seanmil:
Add tests demonstrating the regression introduced in 2.6.0
as described in GH issue #565


rodjek:
Replaced by #582

PR #569, opened 2017-07-16
Fix adapters compatibility for older Puppet 3.x

jearls:
Several of the configuration parameters listed in the Adapter3X class
were not compatible with older versions of Puppet 3.x.  For puppet
modules trying to maintain compatibility with older installations, this
caused rspec to fail.  This change splits the Adapter3X class into
multiple classes for different versions, each building on the previous
to add the configuration parameters introduced in each version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ A hash of default trusted facts that should be used for all the tests
 #### trusted\_node\_data
 Type    | Default | Puppet Version(s)
 ------- | ------- | -----------------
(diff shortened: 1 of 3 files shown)


rodjek:
Thanks for the PR!

Just as a heads up, I'm going to be deprecating Puppet < 4.x support soon as it's EOL (and dropping it entirely in the 3.0.0 release)

rodjek:
2.6.5 has been pushed to rubygems

PR #579, opened 2017-08-02
Add simple pre-release tests

rodjek:
The `release_test` rake task pulls down a selection of real world Puppet
modules and runs them through the following process:

 1. bundle install
 2. bundle exec rake spec (to get a baseline result)
 3. modify gemfile to point to HEAD of local rspec-puppet repo
 4. bundle install
 5. bundle exec rake spec (to get a result for rspec-puppet HEAD)
 6. compare the summaries of the two rspec results and alert if different.

This process is rather time intensive and is not suitable for on-push CI, but
is instead intended to be used before pushing out a new release to increase
confidence in the new release.

The current module list includes some Puppet modules and some community
modules, targetting both Linux and Windows environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /.bundle/
 /vendor/
 .ruby-version
+/tmp/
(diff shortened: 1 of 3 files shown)


DavidS:
Great idea!

Issue #592, opened 2017-08-08
b973772 requires metadata.json for rspec-puppet-init to work


Moncky:
I am just playing with rspec-puppet for the first time have been hitting this error:

```
`➜  sslcerts git:(master) ✗ rspec-puppet-init       
Unable to setup rspec-puppet automatically in a control repo
/home/cscott/.rbenv/versions/2.4.1/lib/ruby/2.4.0/fileutils.rb:1054:in `initialize': No such file or directory @ rb_sysopen - spec/fixtures/manifests/site.pp (Errno::ENOENT)
        from /home/cscott/.rbenv/versions/2.4.1/lib/ruby/2.4.0/fileutils.rb:1054:in `open'
        from /home/cscott/.rbenv/versions/2.4.1/lib/ruby/2.4.0/fileutils.rb:1054:in `rescue in block in touch'
        from /home/cscott/.rbenv/versions/2.4.1/lib/ruby/2.4.0/fileutils.rb:1050:in `block in touch'
        from /home/cscott/.rbenv/versions/2.4.1/lib/ruby/2.4.0/fileutils.rb:1048:in `each'
        from /home/cscott/.rbenv/versions/2.4.1/lib/ruby/2.4.0/fileutils.rb:1048:in `touch'
        from /home/cscott/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/rspec-puppet-2.6.7/lib/rspec-puppet/setup.rb:124:in `safe_touch'
        from /home/cscott/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/rspec-puppet-2.6.7/lib/rspec-puppet/setup.rb:16:in `run'
        from /home/cscott/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/rspec-puppet-2.6.7/bin/rspec-puppet-init:20:in `<top (required)>'
        from /home/cscott/.rbenv/versions/2.4.1/bin/rspec-puppet-init:22:in `load'
        from /home/cscott/.rbenv/versions/2.4.1/bin/rspec-puppet-init:22:in `<main>'`
```

The issue is caused by the change in b973772 which requires a metadata.json file to be present.

PR #593, opened 2017-08-08
Add note to create metadata.json to module before running init script

Moncky:
"Resolves" #592 As the implementation is arguably correct this just adds a note to the README for new users.

Should also be updated on the http://rspec-puppet.com/ home page and tutorials


rodjek:
Thanks @Moncky

PR #603, opened 2017-09-11
(HI-581) load hiera windows support before monkey patching ruby

DavidS:
This works:

```ruby
require 'hiera/util/win32'
it do
  is_expected.to compile
end
```

This does not:

```ruby
it do
  require 'hiera/util/win32'
  is_expected.to compile
end
```

And produces the following error:

```
Failure/Error: require 'hiera/util/win32'

LoadError:
  cannot load such file -- fiddle/import
```

### DIFF
--- a/lib/rspec-puppet/monkey_patches.rb
+++ b/lib/rspec-puppet/monkey_patches.rb
@@ -1,5 +1,12 @@
 require 'pathname'
 
+# Load this library before enabling the monkey-patches to avoid HI-581
+begin
+require 'hiera/util/win32'
+rescue LoadError
+  # ignore this on installs without hiera, e.g. puppet 3 gems
+end
+
 class RSpec::Puppet::EventListener
   def self.example_started(example)
     if rspec3?
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: ruby
 cache: bundler
 bundler_args: --without development
+# workaround bundler/rubygems version mismatch on new trusty images for ruby 1.9
+before_install: gem install bundler -v '1.15.4'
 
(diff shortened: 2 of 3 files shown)


lib/rspec-puppet/monkey_patches.rb
```diff
@@ -1,5 +1,12 @@
 require 'pathname'
 
+# Load this library before enabling the monkey-patches to avoid HI-581
+begin
+require 'hiera/util/win32'
```
Iristyle: Hiera 3 is basically EOL, so this path shouldn't change.

But playing the load order game is a bit like playing roulette.  I'd rather see any mocking of `FILE::ALT_SEPARATOR` removed from `rspec-puppet`. Otherwise, this will be a constant game of whack-a-mole any time code is discovered that correctly uses `FILE::ALT_SEPARATOR` to guard execution on Windows.

rodjek: @Iristyle I agree that having to mock these constants is undesirable, but it is an unfortunate necessity at the moment. Long term I'd like to get rid of them, but for now we have to work with what we've got.

reidmv:
Thanks for merging this!

Is there a known timeframe for when an rspec-puppet release containing this fix will be cut? I have a customer who needs it.

DavidS:
It's already on [rubygems](https://rubygems.org/gems/rspec-puppet/versions/2.6.9) and will be part of the next PDK binary release, expected later this week.

reidmv:
Perfect. Thanks @DavidS!!!

DavidS:
All props to @rodjek for being release-trigger-happy.

PR #618, opened 2017-10-25
WIP: Adds rspec configuration to set default facterdb search paths

logicminds:
* FacterDB 0.4.0 added a new feature that allows the user to
    supply their own custom fact sets.  While this feature alone
    has no affect on rspec-puppet.  Other bolt on tools like
    puppet-debugger and rspec-puppet-facts which rely on facterdb
    and driven by rspec-puppet. This commit allows the user
    to set some facterdb environment variables through the rspec
    configuration setting so that using this new facterdb feature
    is more seemless with rspec-puppet.

  * This adds two new configuration settings
     - facterdb_skip_defaultdb
     - facterdb_search_paths

     Both of these settings are set and unset upon each test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 All notable changes to this project will be documented in this file. This
 project adheres to [Semantic Versioning](http://semver.org/).
+## Unreleased
+ * Adds rspec configuration to set default facterdb search paths
 
 ## [2.6.9]
 
--- a/lib/rspec-puppet.rb
+++ b/lib/rspec-puppet.rb
@@ -48,6 +48,8 @@ RSpec.configure do |c|
   c.add_setting :derive_node_facts_from_nodename, :default => true
   c.add_setting :adapter
   c.add_setting :platform, :default => Puppet::Util::Platform.actual_platform
+  c.add_setting :facterdb_skip_defaultdb, :default => false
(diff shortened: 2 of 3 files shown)


logicminds:
I am putting this up for review as I have a few questions.  

1. I want to set the default facterdb search path to the fixtures/facts directory.  However I am unsure how to get the fixtures directory from the rspec-puppet.rb file.  

2. How and where should I be testing these new configurations?  I put them in the support file, but I don't think they should go there.

DavidS (commented):
Overall I'm fine with these changes. They do need to be rebased and the threads from the questions below need to be tied up.

>     1. I want to set the default facterdb search path to the fixtures/facts directory.  However I am unsure how to get the fixtures directory from the rspec-puppet.rb file.

rspec-puppet should not provide a default here. rspec-puppet-facts and facterdb should take care of defaults. If there are problems with those defaults, they should be fixed there.

>     2. How and where should I be testing these new configurations?  I put them in the support file, but I don't think they should go there.

`spec/unit/config_spec.rb` would not be offensive to me.

>     3. How do I also test the before and after state?

set up multiple examples some of which do not have the new config set but are executed after the ones that have it set. If there was leakage between examples this should show it up.

logicminds:
Because rspec-puppet does not actually require facterdb this code should be in the rspec-puppet-facts instead.  I am closing in favor of that approach.

PR #619, opened 2017-11-03
Adds the ability to spec test an arbitrary string

binford2k:
We are building a web service that will enable rspec-puppet testing of
POSTed Puppet code. The use case is a tutorial that will allow users to
type in a snippet of Puppet code and get immediate validation of whether
they'd completed the exercise correctly.

Example test:

```Ruby
require 'rspec-puppet'

RSpec::configure do |c|
  c.string = "package { 'httpd':
    ensure => present,
  }"
end

describe 'apache', :type => :string do
  let(:node) { 'testhost.example.com' }

  describe 'when called with no parameters on redhat' do
    let (:facts) { { ... } }

    it {
      should contain_package('httpd').with({
        'ensure' => 'present',
      })
    }
  end
end
```

### DIFF
--- a/docs/documentation/configuration/index.md
+++ b/docs/documentation/configuration/index.md
@@ -203,3 +203,12 @@ facts with values that it derives from the node name (specified with
 In some circumstances (e.g. where your nodename/certname is not the same as
 your FQDN), this behaviour is undesirable and can be disabled by changing this
 setting to `false`.
+
+### string
+**Type:** String<br />
+**Default:** `nil`<br />
+**Puppet Version(s):** 2.x, 3.x, 4.x, 5.x
+
+Set this to an arbitrary string of Puppet code and define an example of type
+`:string`. Rather than reading files off disk, `rspec-puppet` will run the tests
+against this string of code.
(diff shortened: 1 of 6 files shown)


rodjek:
Rather than making the input string an RSpec setting, would you mind changing it to be a helper on the example instead? e.g. `let(:code)`, `let(:manifest)` or similar?

binford2k:
Hmm. I'm not opposed to that idea, but how would we be able to invoke a spec programmatically then? We're using this now like so:

``` ruby
  def run_rspec(spec_path, str)
    require 'rspec/core'

    # rspec needs an IO object to write to. We just want it as a string...
    data = StringIO.new
    RSpec::configure do |c|
        c.output_stream = data
        c.formatter     = 'json'
    end

    # require *after* setting the output stream or it screams at us
    # String input depends on https://github.com/rodjek/rspec-puppet/pull/619
    require 'rspec-puppet'
    RSpec::configure do |c|
        c.string        = str
    end

    begin
        RSpec::Core::Runner.run([spec_path])
        return data.string
    rescue StandardError, LoadError => e
        return "Error running spec test: #{e.message}",
end
```

DavidS:
@binford2k @scotje is this still relevant? If yes, please link to a internal ticket to capture that.

sanfrancrisko:
I'm going to close this issue due to lack of activity, although it can be reopened again if needs be.

sanfrancrisko:
@binford2k Do you intend to return to this? If not, would you be OK closing the PR?

binford2k:
@sanfrancrisko I would like to get it merged, but I don't know how to accomplish what @rodjek asked for and retain the ability to run a spec test from an arbitrary file against an arbitrary string.

PR #630, opened 2017-12-21
Reset default providers for Types before compiling catalogue

rodjek:
Previously, if the catalogue had been converted into a RAL catalogue
(done when using the `compile.with_all_deps` matcher), the default
provider would be cached for the duration of the Ruby process and any
future `Puppet::Type` objects would have that provider set, even if the
Facter values had changed and the provider was no longer suitable.

This lead to unpredictable behaviour when switching between compiling
for different platforms and when parallelising the test execution.

The simple solution to this is to clear cached default provider for each
Puppet::Type before compiling the catalogue.

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -337,6 +337,8 @@ module RSpec::Puppet
 
       stub_facts! facts_val
 
+      Puppet::Type.eachtype { |type| type.defaultprovider = nil }
+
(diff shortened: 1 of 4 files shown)


asedge:
Thank you @rodjek!  I'll pull this and #631 and try them out.

asedge:
@rodjek This PR, #631 and #608 fix all the issues we discussed yesterday.  Thanks so much!

PR #637, opened 2018-01-16
(#625) Provide a mock Win32::TaskScheduler for non-windows hosts

### LINKED ISSUE #625: scheduled_task trigger seems to break rspec-puppet - undefined method validate_trigger for nil:NilClass
Performing rspec testing of a puppet class containing a `scheduled_task` resource fails with the following error message if it contains the `trigger` parameter: 

```
failed: rspec: ./spec/classes/sausage_spec.rb:8: error during compilation: Parameter trigger failed on Scheduled_task[foo]: Validate method failed for class trigger: undefined method `validate_trigger' for nil:NilClass at /Users/jesse/src/puppet/tmp/pdk_test/sausage/spec/fixtures/modules/sausage/manifests/init.pp:10
  sausage on windows-2012 R2-x64 should compile into a catalogue without dependency cycles
  Failure/Error:
        let(:facts) { os_facts }

        it { is_expected.to compile }
      end
    end
```

rodjek:
Scheduled_task resources require the constants defined in `Win32::TaskScheduler` in order to validate their parameters. Unfortunately, this library is not usable on non-Windows hosts as it uses FFI to call into kernel32. Instead, we'll include a mock version of the class with the constants defined and alias it to `Win32::TaskScheduler` if it is not available. In order to catch any possible drift between the mock values and the real ones, when the spec is run on windows it will iterate through all the constants in the real `Win32::TaskScheduler` and ensure that the mocked constant is defined properly.

Fixes #625


rodjek:
@asedge another one for you to try out :)

asedge:
Awesome, thanks! I'll let you know.

asedge:
@rodjek This works, thanks!

PR #646, opened 2018-02-02
Fix ability to test powershell exec provider on non-windows os

TraGicCode:
Currently if you create a manifest with an exec that uses the powershell provider the tests blow up when run on a non-windows o/s.  This is because at the top of the powershell exec provider there is code that brings in helpers.  These helpers try and require the win32/registry library which doesnt exist on a non-windows machine.  In order to get around this issue i essentially did what was done for the scheduled task monkey patching.  I've added tests and verified this works and also below is the information on how to reproduce this bug

1. Make your your not on a windows machine ( i'm on a mac ).
2. create a manifest that contains explicitly calls out to use the powershell exec provider

```puppet

exec { 'rename-guest':
  command   => '$(Get-WMIObject Win32_UserAccount -Filter "Name=\'guest\'").Rename("new-guest")',
  unless    => 'if (Get-WmiObject Win32_UserAccount -Filter "Name=\'guest\'") { exit 1 }',
  provider  => powershell,
}
```
3. run a simple rspec compile test against it.


rodjek:
Looks good to me, thanks!

PR #650, opened 2018-02-26
Add puppet-catalog_rspec related project

hajee:


rodjek:
Thanks, I wasn't aware of that project :)

PR #658, opened 2018-03-06
Only prepare server_facts hash if trusted_server_facts is enabled

rodjek:
This fixes up the issue in 2.6.10 that @rnelson0 raised in #599.


trevor-vaughan:
#659 may be relevant to this change.

PR #667, opened 2018-03-22
fix Puppet 3 map/reduce

jeremy-asher:
- (5bdb7e7cda1fd46b8043087201b02f03d24b1b5a) add test for map/reduce functions
  This adds a test to exercise the map and reduce syntax which requires the
  future parser or Puppet 4+.

- (010afc315b2e809c75665169e9de1ab48f446a18) use the correct Lexer for the future parser
  This fixes a bug that occurs when processing certain future parser compliant 
  text on Puppet < 4.0 by switching to the newer lexer when the future parser is 
  enabled.

I'm not sure if the change I made to `spec_helper.rb` is the best route to get the future parser tested (using the `FUTURE_PARSER` environment variable), but this definitely fixes issues parsing newer style code on Puppet 3.  If you still care about Puppet 3 compatibility, it wouldn't hurt to add this combination of environment vars to the automated testing as well.

### DIFF
--- a/lib/rspec-puppet/setup.rb
+++ b/lib/rspec-puppet/setup.rb
@@ -1,7 +1,4 @@
 require 'puppet'
-if Puppet.version.to_f >= 4.0
-  require 'puppet/pops'
-end
 require 'fileutils'
 
 module RSpec::Puppet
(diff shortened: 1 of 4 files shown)


jeremy-asher:
Actually, I think this isn't quite working as I expected due to the timing of settings being loaded before they are checked in `setup.rb`.  The test may also not be quite right.  I'll try to sort this out, but am still curious if this is something that seems reasonable to merge.

jeremy-asher:
Okay, I updated the second commit to delay the conditional import of `puppet/pops` and to base the conditional on `RSpec.configuration` instead of reading the value of `Puppet[:parser]`.  This fixes the issue in my local, manual testing.

The unit test I attempted to write is not demonstrating the issue due to the fixture setup (where this issue exists) being skipped there.  I'd like to fix the test, but I'm not seeing an existing unit test that exercises the fixture setup code to base this test off of and I was somewhat stumped trying to figure out a reasonable way to add one.

PR #675, opened 2018-04-10
(MODULES-6978) only look in 'os' fact if 'os' is a hash

eputnam:
I have been working on a failure in the puppetlabs-inifile unit tests
recently and found out that rspec-puppet may be looking in the wrong
place when setting the 'pretend_platform' for tests. In the line changed, 'os'
should be ommitted and and then handled below where it is instead
evaluated as a hash. So that if you use structured facts in your tests,
it will not look in 'os', find a hash, then default to posix because a
hash is not 'windows', instead of waiting to look at ['os']['family'].


eputnam:
failures appear to be unrelated to my changes

PR #703, opened 2018-06-29
Cite create_specs in related projects section

alexharv074:
I am the author of create_specs, a project that is similar to
puppet-catalog_rspec, although mine is a rather different implementation
and more featureful.

It seems to make sense to mention it here.


rodjek:
Cool!

PR #716, opened 2018-08-20
Improve self test coverage

rodjek:


DavidS:
Rebased to see where this work is standing with current test matrix

PR #718, opened 2018-08-23
Only set :autosign if the setting exists

binford2k:
See https://tickets.puppetlabs.com/browse/PUP-3648 for discussion. This setting is deprecated and will be removed in Puppet 6


binford2k:
Heh. I clearly didn't notice #717

PR #720, opened 2018-09-06
Patch service to default to systemd on Debian 9

hashar:
Puppet 4.8.x systemd provider does not recognize Debian 9 (Stretch), it
thus fall back to runit and the spec fails with:

  Could not find the daemon directory (tested [/etc/sv,/var/lib/service])

Tested using:

 PUPPET_GEM_VERSION='4.8.2' bundle install
 bundle exec rspec ./spec/classes/relationship__titles_spec.rb

References:
https://github.com/puppetlabs/puppet/commit/d5a69fb57c15683873b422cb5e17ef06ca13cea5
https://phabricator.wikimedia.org/T203645

### DIFF
--- a/lib/rspec-puppet/monkey_patches.rb
+++ b/lib/rspec-puppet/monkey_patches.rb
@@ -315,6 +315,18 @@ Puppet::Type.type(:file).provide(:windows).class_eval do
   end
 end
 
+# Puppet 4.8 systemd provider does not recognize Debian 9 (Stretch). It fall
+# back to runit which causes:
+#   Could not find the daemon directory (tested [/etc/sv,/var/lib/service])
+# The Debian package is patched, but the gem 4.8.2 is not.
+# Support has been added in Puppet 4.10.0:
+# https://github.com/puppetlabs/puppet/commit/d5a69fb57c15683873b422cb5e17ef06ca13cea5
+if Puppet::Util::Package.versioncmp(Puppet.version, '4.8.0') >= 0 && Puppet::Util::Package.versioncmp(Puppet.version, '4.10.0') < 0
+  Puppet::Type.type(:service).provide(:systemd).instance_eval do
+    defaultfor :operatingsystem => :debian
+  end
+end
+
 # Prevent Puppet from requiring 'puppet/util/windows' if we're pretending to be
 # windows, otherwise it will require other libraries that probably won't be
 # available on non-windows hosts.
--- a/spec/classes/test_multi_os.rb
+++ b/spec/classes/test_multi_os.rb
@@ -23,5 +23,12 @@ describe 'test::multi_os' do
     it 'sets the provider of the File resource to :posix' do
       catalogue.resource('file', '/test').to_ral.provider.class.name.should eq(:posix)
     end
+
+    describe "service resource" do
+      let (:pre_condition) { 'service { "foo": }' }
+      it 'sets provider to systemd' do
+        catalogue.resource('service', 'foo').to_ral.provider.class.name.should eq(:systemd)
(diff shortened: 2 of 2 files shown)


hashar:
Debian Stretch comes with puppet 4.8.2, the Debian package has been patched but the gem from rubygems.org is not.

A potential fix is to add to every spec_helpers:
```
service_type = Puppet::Type.type(:service)
service_type.provider_hash[:systemd].defaultfor :operatingsystem => :debian
```

I would rather have it handled by rspec-puppet :]

lib/rspec-puppet/monkey_patches.rb
```diff
@@ -315,6 +315,18 @@ def manages_symlinks?
...
 
+# Puppet 4.8 systemd provider does not recognize Debian 9 (Stretch). It fall
+# back to runit which causes:
+#   Could not find the daemon directory (tested [/etc/sv,/var/lib/service])
+# The Debian package is patched, but the gem 4.8.2 is not.
+# Support has been added in Puppet 4.10.0:
+# https://github.com/puppetlabs/puppet/commit/d5a69fb57c15683873b422cb5e17ef06ca13cea5
+if Puppet::Util::Package.versioncmp(Puppet.version, '4.8.0') >= 0 && Puppet::Util::Package.versioncmp(Puppet.version, '4.10.0') < 0
+  Puppet::Type.type(:service).provide(:systemd).instance_eval do
+    defaultfor :operatingsystem => :debian
```
lavagetto: This condition is incorrect, as it will make tests for Debian < 8.0 fail or be inaccurate (those systems don't use systemd).

Also, the condition set in puppet 4.10.x is still wrong and overridden in the debian pacakges to this day.

The correct condition would probably be:
- Monkey-patch all versions that include a systemd provider and offer the defaultfor method for providers
- Add a specific defaultfor to the "debian" provider for debian versions up to 7 (you can limit yourself to 5,6,7)
- Add this defaultfor

hashar: And on further discussion, @lavagetto pointed that the patch modifies the behavior of the puppet interpreter. That could potentially hide an issue when running puppet from Gem, but I guess everyone just use the patched up puppet debian package.

TheMeier:
We see the same problem is there an open issue to this PR?

TheMeier:
The proposed patch does not work for me. If the situation arises where we run into this problem this patch leads to an error:
```
Puppet::Error:
  Could not autoload puppet/type/service: Could not autoload puppet/provider/service/runit: /builds/pocu/xymonclient/vendor/bundle/ruby/2.4.0/gems/puppet-5.5.17/lib/puppet/provider/service/runit.rb:110: syntax error, unexpected end-of-input, expecting keyword_end
```
The same error i run into with the first proposed patch in https://github.com/puppetlabs/puppet/pull/7794. There is an update version there now

DavidS:
I'm very uncomfortable with this change as it is modifying the underlying puppet's behaviour. Since these test results seem to reflect reality (wrong behaviour on debian), please do patch your puppet version to not exhibit buggy behaviour, or upgrade to a version that is actually supported and maintained.

hashar:
The Puppet version shipped by Debian is patched.  The issue lies in the Puppet 4.8 / 4.10 rubygems which, being no more supported at the time, lacked the bit to flag `systemd` as the default.  When running tests with a Gemfile to install Puppet, that leads to the failure.  If I remember properly, we ended up using monkey patching and this pull request is merely to upstream our monkey patch.

Then Stretch / Puppet 4.x are old, so probably we should not bother with this PR anymore.

PR #721, opened 2018-09-25
(MODULES-7854) Stub out default values for puppet6

joshcooper:
Puppet6 modified basemodulepath so it's default value depends on the
env_windows_installdir fact. It also added a new setting,
vendormoduledir, that does the same. This commit stubs out the default
values for these settings as is done with modulepath. This behavior only
occurs when testing against Puppet6, so it's shouldn't affect older
puppet versions which don't define the vendormoduledir setting or use a
fact to determine the default basemodulepath.

See also PUP-8582.

### DIFF
--- a/lib/rspec-puppet.rb
+++ b/lib/rspec-puppet.rb
@@ -52,6 +52,8 @@ RSpec.configure do |c|
   c.add_setting :derive_node_facts_from_nodename, :default => true
   c.add_setting :adapter
   c.add_setting :platform, :default => Puppet::Util::Platform.actual_platform
+  c.add_setting :vendormoduledir, :default => Puppet::Util::Platform.actually_windows? ? 'c:/nul/' : '/dev/null'
+  c.add_setting :basemodulepath, :default => Puppet::Util::Platform.actually_windows? ? 'c:/nul/' : '/dev/null'
 
   c.instance_eval do
     def trusted_server_facts
--- a/lib/rspec-puppet/adapters.rb
+++ b/lib/rspec-puppet/adapters.rb
@@ -192,7 +192,14 @@ module RSpec::Puppet
       end
     end
 
-    class Adapter6X < Adapter40; end
+    class Adapter6X < Adapter40
+      def settings_map
+        super.concat([
+          [:basemodulepath, :basemodulepath],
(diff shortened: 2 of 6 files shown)


joshcooper:
DO NOT MERGE, still testing

joshcooper:
It looks like rspec-pupppet#master fails against puppet#master unrelated to this PR. I filed it as https://tickets.puppetlabs.com/browse/MODULES-7855.

joshcooper:
Still need to address the issue where `Puppet:Util.path_to_uri` gets called and rspec-puppet's monkey patching leaves puppet in an inconsistent state. I created a `foo` module and class using `pdk`

```ruby
require 'spec_helper'

describe 'foo' do
  on_supported_os.each do |os, os_facts|
    context "on #{os}" do
      let(:facts) { os_facts }

      it { is_expected.to compile }
    end
  end
end
```

When I run the following the `windows?` and `microsoft_windows?` don't match, causing `path_to_uri` to fail with a path of the form `C:/foo/.../component.rb`, because the code for [prepending a leading slash to the windows path](https://github.com/puppetlabs/puppet/blob/6.0.0/lib/puppet/util.rb#L331) is skipped.
 
```
C:\foo>bundle exec rspec spec
...
foo
  on centos-7-x86_64
** ALT_SEPARATOR
** Puppet::Util::Platform.windows? true
** Puppet.features.microsoft_windows? true
** ALT_SEPARATOR \
** Puppet::Util::Platform.windows? false
** Puppet.features.microsoft_windows? true
    should compile into a catalogue without dependency cycles (FAILED - 1)
```

The first time `path_to_uri` is called, everything is consistent. But then it's called again seemingly after we're switched back to Windows.

davidmalloncares:
Hey @joshcooper / @rodjek - do you guys have a rough idea on timeline for this fix?

joshcooper:
@davidmalloncares never closer, hopefully very soon.

joshcooper:
Travis passed, but shows as pending... not sure what's up there. Looks good to me @rodjek

rodjek:
Super close! Just poking the timed out tests (it happens sometimes) so that we can have a nice green merge. Thanks for the help @joshcooper!

PR #724, opened 2018-09-28
(MODULES-7855) Fix test issues caused by new Puppet 6 validation.

kris-bosland:
Puppet 6 now validates expected module top level constructs and illegal definition locations.  This update fixes rspec-puppet test issues related to those new validation errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@
 /vendor/
 .ruby-version
 /tmp/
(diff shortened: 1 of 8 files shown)


kris-bosland:
Note that we are deleting an empty init.pp file, since it is not needed for the tests and as of Puppet 6 triggers an error for an illegal top level construct (the validator complains about a "Nop Expression").

PR #730, opened 2018-10-06
docs : add note about rspec-puppet-facts

Dan33l:
I know it is `rspec-puppet` doc but it should be interesting to get a note about `rspec-puppet-facts` in facts testing section.


rodjek:
Good idea, thanks @Dan33l!

Dan33l:
Nice rework. Thank you.

PR #731, opened 2018-10-23
Revert "Reencode strings when matching to update line endings"

rodjek:
Reverts rodjek/rspec-puppet#726

### DIFF
--- a/lib/rspec-puppet/matchers/parameter_matcher.rb
+++ b/lib/rspec-puppet/matchers/parameter_matcher.rb
@@ -78,7 +78,7 @@ module RSpec::Puppet
       end
 
       def check_regexp(expected, actual)
-        !!(munge_line_endings(actual.to_s).match expected) == @should_match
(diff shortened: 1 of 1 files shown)


michaeltlombardi:
It looks like the appveyor builds are failing [here](https://github.com/rodjek/rspec-puppet/blob/master/spec/unit/monkey_patches/windows/taskschedulerconstants_spec.rb#L3):

```
./spec/unit/monkey_patches/windows/taskschedulerconstants_spec.rb:3
```

This seems to be related to some [monkey patching](https://github.com/rodjek/rspec-puppet/blob/master/lib/rspec-puppet/monkey_patches/windows/taskschedulerconstants.rb) performed in rspec-puppet on the previously included taskscheduler type/provider inside of puppet - that has since been moved out into it's [own module](https://github.com/puppetlabs/puppetlabs-scheduled_task/) as of puppet 6.0.0.

I'm not sure what the monkey patching was being used for initially, but maybe they're no longer required?

Issue #626, opened 2017-12-07
`hiera_config` doesn't override puppet 5 layer 3 `hiera.yaml`


LongLiveCHIEF:
As of puppet 5, hiera is now native to puppet. When a module as `hiera.yaml` at it's root, it's given precedence over any values passed in your `spec/spec_helper.rb` for `hiera_config`.

As an example, my module's `hiera.yaml` hierarchy contains a eyaml data source. This has caused some issues ([documented here](https://github.com/voxpupuli/hiera-eyaml/issues/253)) which I tried to get around by providing rspec-puppet a different hiera.yaml file to use for testing, that didn't contain the eyaml data sources. 

Unfortunately, it appears that even though this value is passed, it doesn't trump the layer 3 hiera specification of the module's own `hiera.yaml` file. Even though i've confirmed that rspec-puppet loads the config file specified in my `spec/spec_helper.rb`, it does not use values from the data sources in that configs hierarchy. It uses the data-sources defined in the module's own `hiera.yaml` file.

Is there a way we can ensure that the following overrides the layer 3 `hiera.yaml` file in puppet 5 modules?

```
require 'puppetlabs_spec_helper/module_spec_helper'
require 'rspec-puppet-facts'
include RspecPuppetFacts

fixture_path = File.expand_path(File.join(__FILE__, '..', 'fixutes'))

RSpec.configure do |c|
  c.default_facts = default_facts
  c.hiera_config = File.join(fixture_path, 'hiera.yaml')
end
```

fabx:
+1

rodjek:
@LongLiveCHIEF Can you create a simple repo with failing tests that show your expected behavior here? I had a look at the repo you provided in the linked issue but there were no failing tests in there. I'm not entirely sure what you're trying to do here (hiera 5 is a very different beast to hiera 3) and some failing examples will clarify things for me :)

nabertrand:
I added the following to `spec/spec_helper.rb` to cause Puppet to use a `hiera.yaml` from the module `spec` folder if found. The code could definitely be extended to support custom `hiera.yaml` locations for each module, but this got me around the eyaml issue. I decided to search the spec folder instead of hard-coding a path as there doesn't seem to be a standard location for spec `hiera.yaml` files. For example, some sites suggest `spec/fixtures/hiera/hiera.yaml`, whereas others suggest `spec/fixtures/hiera.yaml`.

```ruby
module Puppet
[...]

PR #732, opened 2018-10-30
Allow customizing module-layer Hiera configuration

nabertrand:
This adds four new RSpec configuration parameters to modify the behavior of module-layer Hiera data.

- disable_module_hiera
  - Enabling this will prevent Puppet from using module-layer Hiera data entirely. This includes the module being tested as well as any fixture modules. The end effect is that only Hiera data from the global :hiera_config parameter will be used.
- use_fixture_spec_hiera
  - Enabling this will prevent Puppet from using the module-layer Hiera config file and instead search the module spec folder for a file named hiera.yaml.
- fallback_to_default_hiera
  - Controls whether or not to fall back to the default hiera.yaml file for the module layer if use_fixture_spec_hiera is enabled and no spec hiera.yaml file is found. Defaults to true.
- fixture_hiera_configs
  - A hash of module names and their respective module-layer Hiera config file paths. This can be used to override the path to the module-layer hiera.yaml.

The original impetus for creating these settings was to solve errors generated from modules that had eyaml enabled in their hiera config files. Even with :hiera_config set appropriately, Puppet would still try to load hiera data from the module layer. See #626.

If these settings seem useful for merging, I would be glad to update any appropriate documentation.

### DIFF
--- a/lib/rspec-puppet.rb
+++ b/lib/rspec-puppet.rb
@@ -54,6 +54,10 @@ RSpec.configure do |c|
   c.add_setting :platform, :default => Puppet::Util::Platform.actual_platform
   c.add_setting :vendormoduledir, :default => Puppet::Util::Platform.actually_windows? ? 'c:/nul/' : '/dev/null'
   c.add_setting :basemodulepath, :default => Puppet::Util::Platform.actually_windows? ? 'c:/nul/' : '/dev/null'
+  c.add_setting :disable_module_hiera, :default => false
+  c.add_setting :fixture_hiera_configs, :default => {}
+  c.add_setting :use_fixture_spec_hiera, :default => false
+  c.add_setting :fallback_to_default_hiera, :default => true
 
   c.instance_eval do
     def trusted_server_facts
--- a/lib/rspec-puppet/monkey_patches.rb
+++ b/lib/rspec-puppet/monkey_patches.rb
@@ -258,6 +258,40 @@ module Puppet
       end
     end
   end
+
+  if Puppet::Util::Package.versioncmp(Puppet.version, '4.9.0') >= 0
+    class Module
+      old_hiera_conf_file = instance_method(:hiera_conf_file)
+      define_method(:hiera_conf_file) do
+        if RSpec::Puppet.rspec_puppet_example?
+          if RSpec.configuration.disable_module_hiera
+            return nil
+          elsif RSpec.configuration.fixture_hiera_configs.key?(name)
+            config = RSpec.configuration.fixture_hiera_configs[name]
+            config = File.absolute_path(config, path) unless config.nil?
+            return config
+          elsif RSpec.configuration.use_fixture_spec_hiera
+            config = RSpec::Puppet.current_example.fixture_spec_hiera_conf(self)
+            return config unless config.nil? && RSpec.configuration.fallback_to_default_hiera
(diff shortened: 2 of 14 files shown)


vStone:
Would it be possible to have use the fixtures/hiera.yaml and module hiera.yaml at the same time? 

It would feel more natural if the fixtures/hiera.yaml acts as global or environment level hiera configuration

nabertrand:
> Would it be possible to have use the fixtures/hiera.yaml and module hiera.yaml at the same time?
> 
> It would feel more natural if the fixtures/hiera.yaml acts as global or environment level hiera configuration

I find myself easily confused when trying to describe the different hiera configs and layers involved when spec-testing a module. Perhaps a directory tree might help me better understand the concern, where `test-module` is the module being tested by rspec and `fixture-module` is a test dependency of `test-module`:
```
test-module/hiera.yaml
test-module/spec/fixtures/hiera.yaml
test-module/spec/fixtures/modules/fixture-module/hiera.yaml
test-module/spec/fixtures/modules/fixture-module/data/common.yaml
test-module/spec/fixtures/modules/fixture-module/spec/fixtures/hiera.yaml
test-module/spec/fixtures/modules/fixture-module/spec/fixtures/data/common.yaml
```
If you're asking about using `test-module/hiera.yaml` as the module-layer hiera config for `test-module` and `test-module/spec/fixtures/hiera.yaml` as the global-layer hiera config, then I believe that is how rspec-puppet always behaved by setting `:hiera_config` to `spec/fixtures/hiera.yaml`. Without this PR, setting `:hiera_config` has no effect on the module layer so Puppet continues to use the default location for module-layer hiera configs of each module, e.g. `test-module/spec/fixtures/modules/fixture-module/hiera.yaml` would be the module-layer hiera config location for `fixture-module`.

If you're asking about using `test-module/spec/fixtures/modules/fixture-module/hiera.yaml` as the module-layer config for `fixture-module` and still being able to have `test-module/spec/fixtures/modules/fixture-module/spec/fixtures/hiera.yaml` supercede values set in `test-module/spec/fixtures/modules/fixture-module/hiera.yaml`, the easiest solution I can see is referencing the module-layer hiera data in `test-module/spec/fixtures/modules/fixture-module/spec/fixtures/hiera.yaml`:
```yaml
---
￼	version: 5
￼	defaults:
￼	  data_hash: yaml_data
￼	  datadir: data
￼	hierarchy:
￼	  - name: Production module data
￼	    path: ../../../data/common.yaml
￼	  - name: RSpec module data
￼	    path: common.yaml
```
Does that answer your question?

rodjek:
@nabertrand Thanks for doing this. I'm more than happy to merge this in once the documentation has been updated. The new settings definitely need to be added to the configuration reference https://github.com/rodjek/rspec-puppet/blob/master/docs/documentation/configuration/index.md.

As you mentioned though, the interaction of the different hiera layers with spec testing can be a confusing subject, so maybe a bit of a long form explanation as an fragment in https://github.com/rodjek/rspec-puppet/tree/master/docs/_includes that can then be included in the "testing classes", "testing defined types" pages as well.

sanfrancrisko:
https://github.com/rodjek/rspec-puppet/pull/809 encompasses all of this work, in addition to updating the configuration documentation and resolving the merge conflicts on the branch.

PR #734, opened 2018-11-11
Fix a typo

pegasd:


pegasd:
I really doubt the tests are failing because of the change.
Is there something going on right now in the appveyor domain that leads to failing tests?

PR #747, opened 2019-02-27
Add matcher for detecting duplicate attributes

binford2k:
This allows writing specs like:

``` ruby
it {
  should not_duplicate_attribute('user', 'uid')
}
```

The use case is validating that specified type/attribute pairs are
unique. For example, if a manifest had a large number of user
declarations, you'd want the ability to ensure that the resulting
catalog didn't contain any duplicated UIDs.


binford2k:
@reidmv what do you think of this?

@mikkergimenez could you describe your use case a little better than I did?

mikkergimenez:
This is exactly what I'm trying to do.

DavidS:
I've incorporated all the comments from this PR and added tests and docs. This should be ready for final sign-off and merge.

scotje (approved):
Looks good to me!

PR #748, opened 2019-03-05
Support directories for the manifest setting

tdevelioglu:
Rspec-puppet currently doesn't support loading a directory as a manifest
like the `manifest` setting in Puppet does.

When testing environments that have a directory as a manifest you have
to resort to manually loading all the files in the directory in a
:pre_condition which leads to a clumsy experience (especially when you
need :pre_condition for other reasons).

This commit should solve this by alphabetically loading all .pp files if
manifest is a directory.

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -129,13 +129,18 @@ module RSpec::Puppet
     end
 
     def site_pp_str
-      site_pp_str = ''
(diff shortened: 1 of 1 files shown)


pegasd:
I believe the “else” is way too generic here. There are other file types besides a directory and a regular file - symlinks, device files and others.

tdevelioglu:
@pegasd I've amended the commit.

PR #755, opened 2019-06-07
Revert "Support directories for the manifest setting"

npwalker:
Reverts rodjek/rspec-puppet#748

This is a breaking change.  For our modules we currently place various .pp files in `spec/fixtures/manifests`.  Many of those .pp files conflict with existing pre_condition statements we have so after this change when in all of our tests failed with duplicate declaration errors.

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -129,18 +129,13 @@ module RSpec::Puppet
     end
 
     def site_pp_str
-      return '' unless (path = adapter.manifest)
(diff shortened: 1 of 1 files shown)


jpartlow:
@npwalker this can be closed (I don't have permission); I've removed the duplicate class declaration from puppet_enterprise and unpinned rspec-puppet (https://github.com/puppetlabs/puppet-enterprise-modules/commit/fac9174e5e2af0db91749b0f5a477154d4b01e9a).

PR #767, opened 2019-10-06
Fix support for `rspec-expectations` 3.8.5+

pirj:
This change adds value expectation support to RSpec's built-in `raise_error` matcher that is quite often used in projects using `rspec-puppet`, e.g.:

    subject(:catalogue) { lambda { load_catalogue } }
    it { is_expected.to raise_error Puppet::Error }

Also, since block matchers are more strict by default and don't accept value targets (defined as lambdas), this change adds an indicator that built-in `rspec-puppet` matchers do support value expectations as well as block expectations. This means that both:

    it { expect(catalogue).to contain_file('README') }
    it { expect { catalogue }.to contain_file('README') }

will work.

I'm not 100% certain this will work. Tested with https://github.com/voxpupuli/puppet-cassandra.
Appreciate if you can point out.

cc @alexjfisher @cyberious do you mind testing this against some of the projects depending on `rspec-puppet` that you maintain?

Related:
RSpec issue regarding the removal of implicit block syntax in a patch version https://github.com/rspec/rspec-expectations/issues/1134
RSpec pull request with the removal https://github.com/rspec/rspec-expectations/pull/1125
https://blog.rubystyle.guide/rspec/2019/07/17/rspec-implicit-block-syntax.html
https://github.com/rubocop-hq/rspec-style-guide/issues/76


rodjek:
Thanks so much @pirj! I've run this change against a bunch of modules and it looks good :)

pirj:
Awesome, thank you!

PR #784, opened 2020-02-16
monkey-patch match_manifests to not load init.pp when not necessary

kjetilho:
When loading a class, Puppet::Module::match_manifests is used to list
candidate filenames.  To maintain backwards compatibility, it will return
init.pp for the module first in the list, so that a class may be declared
outside autoload namespace (Puppet bug #4220).

Normally, this does not cause any problems, but if pre_condition is used
to make a simplified version of the top class in your module, Puppet will
complain about a duplicate declaration.

```ruby
  describe "baseconfig::logging" do
    let(:pre_condition) { [
     'class baseconfig($enable_filebeat=false) { }',
     'include baseconfig',
    ] }
```

This results in an error like this:

  error during compilation: Class 'baseconfig' is already defined (line: 2); cannot
  redefine (file: /home/kjetilho/src/baseconfig/spec/fixtures/modules/baseconfig/manifests/init.pp,
  line: 6) on node ranger.ms.redpill-linpro.com

The patch changes the logic so that it removes init.pp from the beginning
of the list if there are other files in the list.  This will break code
which puts the class foo::bar in foo/manifests/init.pp *and* has a file
foo/manifests/bar.pp in the project, but I believe such code thoroughly
deserves to break.

### DIFF
--- a/lib/rspec-puppet/monkey_patches.rb
+++ b/lib/rspec-puppet/monkey_patches.rb
@@ -297,12 +297,12 @@ end
 # "mod::foo", this causes duplicate declaration for "mod".
 # This monkey patch only loads "init.pp" if "foo.pp" does not exist.
 class Puppet::Module
-  if instance_methods.include?(:match_manifests)
+  if [:match_manifests, 'match_manifests'].any? { |r| instance_methods.include?(r) }
     old_match_manifests = instance_method(:match_manifests)
 
     define_method(:match_manifests) do |rest|
       result = old_match_manifests.bind(self).call(rest)
-      if result.length > 1 && result[0] =~ %r{/init.pp$}
+      if result.length > 1 && File.basename(result[0]) == 'init.pp'
         result.shift
       end
       result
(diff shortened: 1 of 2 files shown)


kjetilho:
the Travis CI failure seems to be a problem in the bleeding edge Puppet gem, not related to my patch.  how can I rerun the check?

rodjek:
@kjetilho Can you add a test case for this?

kjetilho:
hmm, not sure how to do this.  do you mean the basic Ruby behaviour of the method (in spec/unit/monkey_patches_spec.rb) or the puppet test cases?

I will note that

* `include sysctl::common` serves as a test that my patch doesn't break existing behaviour (as the class is defined in ./spec/fixtures/modules/sysctl/manifests/init.pp)
* ./spec/fixtures/modules/escape/manifests/def.pp can be used to construct a breaking test with existing code:
```ruby
  describe 'escape::def' do
    let (:pre_condition) do [ "class escape() {}" ] end
  end
```
I assume that you will prefer a new "module" for specific tests anyhow?

kjetilho:
ah, you cherry-picked it, thanks!  well, there's my suggested test case for it, I guess you can cherry-pick that, too, if it is the kind of testing you wanted.

Issue #789, opened 2020-04-10
can't modify frozen String: "false"


damluk:
On ubuntu 20.04, ruby 2.7 I get errors all over the place similar to the following:
```
     FrozenError:
       can't modify frozen String: "false"
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:443:in `gsub!'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:443:in `escape_special_chars'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:343:in `str_from_value'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:351:in `block in param_str_from_hash'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:350:in `each'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:350:in `collect'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:350:in `param_str_from_hash'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:280:in `param_str'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:155:in `test_manifest'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:21:in `build_code'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:91:in `block in load_catalogue'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:376:in `with_vardir'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/support.rb:83:in `load_catalogue'
     # ./vendor/ruby/2.7.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/example/class_example_group.rb:7:in `catalogue'
[...]

PR #791, opened 2020-04-10
Run tests on ruby 2.7 successfully

damluk:
Solves #789

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ before_install: gem install bundler -v '1.15.4'
 
 rvm:
   - 2.5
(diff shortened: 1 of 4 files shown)


damluk:
Coverage decrease due to code removal is a false positive (negative?), in my opinion.

Also it seems that AppVeyor does not provide Ruby27-x64 yet.

ekohl:
It would be nice if this could make it into a release

raphink:
This made it into 2.7.9 but was withdrawn in 2.7.10, so 2.7.10 is broken again...

PR #792, opened 2020-04-14
Do not raise failure if nothing is tested

Felixoid:
`RSpec::Puppet::Coverage.report!(percents)` raises test failure if the total amount of tested puppet resources is 0. So either it annoys during ruby code development or should be mitigated with the following snippet in spec_helper.rb

```ruby
RSpec.configure do |c|
  c.after(:suite) do
    RSpec::Puppet::Coverage.report!(94) unless RSpec::Puppet::Coverage.results[:total].zero?
  end
end
```

This PR implement checking of coverage only on total checked puppet resources greater than 0.


scotje:
@Felixoid I'm wondering if #805 addressed this a different way now? (I think `report[:coverage]` should now be `100.00` if there are zero resources)

DavidS:
Closing this as it has been addressed in a different way through #805. Please feel free to re-open this, or create a new issue if your problem persisted.

Felixoid:
THank you. I'm not able to check it out yet but looks pretty promising. I'll come back in case of problems!

PR #793, opened 2020-04-27
Add support for trusted external data

ekinanp:
Signed-off-by: Enis Inan <enis.inan@puppet.com>

### DIFF
--- a/lib/rspec-puppet/support.rb
+++ b/lib/rspec-puppet/support.rb
@@ -88,9 +88,20 @@ module RSpec::Puppet
         hiera_data_value = self.respond_to?(:hiera_data) ? hiera_data : nil
 
         build_facts = facts_hash(node_name)
-        catalogue = build_catalog(node_name, build_facts, trusted_facts_hash(node_name), hiera_config_value,
-                                  build_code(type, manifest_opts), exported, node_params_hash, hiera_data_value,
-                                  RSpec.configuration.trusted_server_facts)
+        catalogue = build_catalog(
+          nodename: node_name,
+          facts_val: build_facts,
+          trusted_facts_val: trusted_facts_hash(node_name),
+          hiera_config_val: hiera_config_value,
+          code: build_code(type, manifest_opts),
+          exported: exported,
+          node_params: node_params_hash,
+          trusted_external_data: trusted_external_data_hash,
+          ignored_cache_params: {
+            hiera_data_value: hiera_data_value,
+            trusted_server_facts: RSpec.configuration.trusted_server_facts,
+          },
+        )
 
         test_module = type == :host ? nil : class_name.split('::').first
         if type == :define
@@ -294,6 +305,19 @@ module RSpec::Puppet
       extensions
     end
 
+    def trusted_external_data_hash
+      return {} unless Puppet::Util::Package.versioncmp(Puppet.version, '6.14.0') >= 0
+
+      external_data = {}
+
+      if RSpec.configuration.default_trusted_external_data.any?
+        external_data.merge!(munge_facts(RSpec.configuration.default_trusted_external_data))
+      end
+
+      external_data.merge!(munge_facts(trusted_external_data)) if self.respond_to?(:trusted_external_data)
+      external_data
+    end
+
     def server_facts_hash
       server_facts = {}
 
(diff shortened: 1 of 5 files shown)


ekinanp:
You can do `$trusted['external']` to access external trusted data. This PR extends rspec-puppet to support mocking that.

According to the [docs](https://puppet.com/docs/puppet/latest/release_notes_puppet.html#experimental-feature:-catalog-compilation-with-external-trusted-data-from-third-parties), this is an experimental feature. However, the "experimental" part is its implementation. The basic API of accessing the data via `$trusted['external']` will remain unchanged.

ekinanp:
Also, implementation's mostly symmetric with https://github.com/rodjek/rspec-puppet/commit/79149532aa1e89097739aed7b3d8560b3f2ad945#diff-71602a9b90b90213fafc0ce1b38e8664

lib/rspec-puppet/support.rb
```diff
@@ -380,7 +393,7 @@ def with_vardir
       end
     end
 
-    def build_catalog_without_cache(nodename, facts_val, trusted_facts_val, hiera_config_val, code, exported, node_params, *_)
+    def build_catalog_without_cache(nodename, facts_val, trusted_facts_val, hiera_config_val, code, exported, node_params, *args)
```
ekinanp, line 396: I can move `external_trusted_data` to come after `trusted_facts_val` instead. Just wasn't sure if doing so would break existing users of this gem and if that is an issue.

ekinanp:
Sounds good. We should change future docs to reference this as trusted_external instead of external_trusted b/c the docs was where I was going off of when naming this stuff.

ekinanp:
@scotje or @DavidS could this get a review? It's needed to properly unit test any manifest that uses `trusted['external']` like e.g. the [ServiceNow CMDB integration](https://github.com/puppetlabs/puppetlabs-servicenow_cmdb_integration)

garrettrowell:
Been using the code in this PR and everything seems to work as expected. Would be great to see this in an upcoming release

lib/rspec-puppet/support.rb
scotje: I guess since `build_catalog` just passes through its `args` splat, changing the signature of this method would effectively change the signature of `build_catalog` as well...

lib/rspec-puppet/support.rb
```diff
@@ -402,10 +415,15 @@ def build_catalog_without_cache(nodename, facts_val, trusted_facts_val, hiera_co
 
       node_obj = Puppet::Node.new(nodename, { :parameters => node_params, :facts => node_facts })
 
+      trusted_info = ['remote', nodename, trusted_facts_val]
+      if Puppet::Util::Package.versioncmp(Puppet.version, '6.14.0') >= 0
+        trusted_external_data = args[-1]
```
DavidS, line 420: This here is more concerning. The args splat's main use was it's side-effect to act as a cache key. This now hard-codes "the last element is the trusted_external data", which is tenuous at best.

Considering that the original use of the build_catalog splat was it's use as cache key, there's the theoretical chance that someone somewhere has (mis-)used this API for their own reasons.

Being saddled with having to support everything back to the stone-age, there's no really good solutions here. One idea would be to start towards cleaning this up for a 3.0 release and change it something like this:

```ruby
def build_catalog_without_cache(nodename, facts_val, trusted_facts_val, hiera_config_val, code, exported, node_params, *_)
  build_catalog_without_cache_new({
    nodename: nodename,
    facts_val: facts_val,
    trusted_facts_val: trusted_facts_val,
    hiera_config_val: hiera_config_val,
    code: code,
    exported: exported,
    node_params: node_params,
    trusted_external: {},
  })
end

def build_catalog(*args)
  @@cache.get(*args) do |*args|
    if args.length == 1
      build_catalog_without_cache_new(args.first)
    else
      build_catalog_without_cache(*args)
    end
  end
end

def load_catalogue(type, exported = false, manifest_opts = {})
  # ...
  catalogue = build_catalog({
    nodename: node_name,
    facts_val: build_facts,
    trusted_facts_val: trusted_facts_hash(node_name),
    hiera_config_val: hiera_config_value,
    code: build_code(type, manifest_opts),
    exported: exported,
    node_params: node_params_hash,
    trusted_external: {},
    ignored_cache_params: {
      hiera_data_value: hiera_data_value,
      trusted_server_facts: RSpec.configuration.trusted_server_facts
    }
  })
  # ...
```

This adds a new 1-argument overload for `build_catalog` that should be a net-new API addition (and thus OK for a .y release) and a net-new `build_catalog_without_cache_new` method that can use the has (and later kwargs) for more flexibility without painting us further into a corner.

DavidS (changes requested):

lib/rspec-puppet/support.rb
ekinanp: Updated a while ago but am wondering why Travis is taking so long...

DavidS (approved):

lib/rspec-puppet/support.rb
DavidS: Thanks for the quick response!

re travis: That's what acquisitions do to services :-(

Issue #796, opened 2020-07-14
rspec-puppet on puppet 6.17 breaks due to removal of default_env


KeithWard:
It looks like default_env was removed in puppet 6.17 - specifically in this commit:

https://github.com/puppetlabs/puppet/commit/94df3c1a3992d89b2d7d5db8a70373c135bdd86b

This results in an exceptiion similar to the following when rspec tries to run tests:
```
 Failure/Error: require 'puppetlabs_spec_helper/module_spec_helper'
 NameError:
   undefined local variable or method `default_env' for Puppet::Util:Module
 # ./vendor/bundle/ruby/2.5.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet/monkey_patches.rb:139:in `get_env'
 # ./vendor/bundle/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/node/environment.rb:501:in `extralibs'
 # ./vendor/bundle/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/node/environment.rb:76:in `initialize'
 # ./vendor/bundle/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/node/environment.rb:43:in `new'
 # ./vendor/bundle/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/node/environment.rb:43:in `create'
 # ./vendor/bundle/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet.rb:240:in `base_context'
 # ./vendor/bundle/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/test/test_helper.rb:53:in `initialize'
 # ./vendor/bundle/ruby/2.5.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet.rb:77:in `block in <top (required)>'
 # ./vendor/bundle/ruby/2.5.0/gems/rspec-puppet-2.7.8/lib/rspec-puppet.rb:34:in `<top (required)>'
 # ./vendor/bundle/ruby/2.5.0/gems/puppetlabs_spec_helper-2.15.0/lib/puppetlabs_spec_helper/module_spec_helper.rb:1:in `require'
 # ./vendor/bundle/ruby/2.5.0/gems/puppetlabs_spec_helper-2.15.0/lib/puppetlabs_spec_helper/module_spec_helper.rb:1:in `<top (required)>'
 # ./spec/spec_helper.rb:7:in `require'
[...]

PR #797, opened 2020-07-14
796 Fix issues with `puppet` gem removal of `default_env`

sgnn7:
Since this is now changed to `DEFAULT_ENV`, we now create the old
function alias to retain functionality.

Bug: https://github.com/rodjek/rspec-puppet/issues/796
Upstream: https://github.com/puppetlabs/puppet/blob/94df3c1a3992d89b2d7d5db8a70373c135bdd86b/lib/puppet/util.rb


scotje:
Tested this locally and it allows `pdk test unit` to run successfully on my test module.

PR #802, opened 2020-08-19
coverage.rb: Fix handling of 0 total resources

raemer:
If there are 0 total resources, a percentage of NaN was generated as
division by zero was not prevented. This checks if the divisor is zero
and sets the percentage to 100 if so.


raemer:
fixes #803

daianamezdrea:
Hi @raemer, can you explain me a little bit the context of this change? I know in some cases a limit of coverage is imposed, is the same situation here ?

scotje:
Hi @raemer, thanks for your contribution! I've refactored this slightly and added a unit test in #805. Once CI passes there I will merge that PR in favor of this one.

PR #807, opened 2020-10-20
Implement RSpec::Puppet::Coverage.add_filter_regex

### LINKED ISSUE #158: [Coverage] Exclude certain resources from coverage report
Would be cool to exclude certain resources ( like anchor's ) from the coverage reports as sometimes these can't be tested.

scotje:
Closes #158

### DIFF
--- a/docs/documentation/coverage/index.md
+++ b/docs/documentation/coverage/index.md
@@ -65,3 +65,26 @@ the required information in them to determine which manifest they came from and
 so can not be excluded from the coverage report.
 </div>
 </div>
+
+To exclude other resources from the coverage reports, for example to avoid `anchor`s,
+use the `add_filter(type, title)` and `add_filter_regex(type, regex)` methods:
+
+{% highlight ruby %}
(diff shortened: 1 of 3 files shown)


scotje:
Only CI failure is due to the old Ruby 1.8-compatible version of rspec not having the `output` matcher.

DavidS (approved):
Added initial shot at docs (please verify!) and added some more handling of edge-cases when processing regex-anchors in filters.

DavidS:
Damnit - ruby 1.8.7 is not supporting `(?>!...)` negative lookbehinds. how about we just drop support for that ruby version?

scotje:
> Damnit - ruby 1.8.7 is not supporting `(?>!...)` negative lookbehinds. how about we just drop support for that ruby version?

Yeah I'm not worried about Ruby 1.8.7 failures, and I'd be OK with updating the gemspec to require at least Ruby 2.1 for the 2.8.0 release.

PR #812, opened 2020-11-03
Add missing links for keep-a-changelog-ish diffs

op-ct:
Looks like they stopped after v6.5.0


DavidS (approved):
appveyor failure because of the (already fixed) `raise_error` issue. Not bothering to re-run for a docs fix.

PR #821, opened 2021-03-15
Add a transitive test property

ekohl:
This is a minimal reproducer that's essentially A ~> B ~> C. The that_subscribes_to matcher shows that this is equal to A ~> C but in practice Puppet doesn't work that way. The service is not restarted if the file is modified.

I don't know what is the expected behavior here, but it does mean that today it's not possible to test for real world behavior.

I've submitted this as a PR since it's the easiest way to submit code and show CI results for it.


sanfrancrisko:
Thanks for highlighting this @ekohl. Let me run this past a few colleagues, but it would seem logical that the test framework is behaving the same way as Puppet behaves, at least. However, we can start a separate conversation with the Puppet core team if you feel there's an issue with how chained notifies are working at the moment?

ekohl:
I would like a formal statement from the core team how they think it should work. From there we can decide how to approach this.

sanfrancrisko:
I pinged some of my colleagues from the agent team and they believe that `A` should notify both `B` and `C` (`C` being triggered by `B`) in the case outlined in https://github.com/theforeman/puppet-foreman_proxy/pull/653. I'll relay your manifest [here](https://github.com/theforeman/puppet-foreman_proxy/pull/653#issuecomment-799398042) to them and see if I can reproduce it too. If so, we'll get a bug raised as it seems to be contradicting the expected behaviour: https://puppet.com/docs/puppet/7.6/lang_relationships.html#lang_rel_chaining_arrows

evgeni:
@sanfrancrisko was there ever any progress on this discussion? We've just hit this problem again and while re-reading https://puppet.com/docs/puppet/7/lang_relationships.html#lang_rel_chaining_arrows I can't really tell if `A ~> B ~> C` should be equivalent to `A ~> B, A ~> C, B ~> C` or `A ~> B, B ~> C`?

More explicitly, the chaining example says "Resource declarations can be chained. That means you can use chaining arrows to make Puppet apply a section of code in the order that it is written. This example applies configuration to the package, the file, and the service, in that order, with **each related resource notifying the next of any changes**" (emph. mine) where one could interpret "any changes" as "any changes to itself" and not "any changes in the whole chain".

ekohl:
Please continue the discussion in https://github.com/puppetlabs/rspec-puppet/pull/29

PR #830, opened 2021-06-08
Add Ruby 3 support

ekohl:


bastelfreak:
@binford2k would it also be possible to get this merged+released or forked to the Puppetlabs namespace?

hashar:
I came here after upgrading Debian to bookworm which comes with ruby3.1. It seems the rodjek/rspec-puppet has been idling since 2021 and the gem is now maintained at https://github.com/puppetlabs/rspec-puppet (which is the repo from where this pull request come).

The commit has been released with rspec-puppet 2.10.0:
```
$ git tag --contains c8e68b53cdae3512f4f0ac3cc1e73fc28511c498
v2.10.0
```

